PR #10, opened 2024-01-29
README cleanup

crandmck:
@gpeacock QUESTION: 

Should we say to do this if you tried before unsuccessfully to install ?
```
pip install --upgrade --force-reinstall c2pa-python
```


gpeacock:
The --upgrade command will install the latest version. The --force_reinstall will remove older versions and install the latest version. So you probably don't need both. I usually just run --force_reinstall since it will ensure the latest is there and also remove older, possibly buggy versions.

PR #15, opened 2024-04-24
New API model with Builder and Reader

gpeacock:
This is the new cross platform API built on streams.
The Builder class is used to construct new manifests.
You can add other ingredients and resources (like thumbnails) directly to the Builder.
The Builder can be saved to an archive and restored for continued work.
The signer now keeps the private key out of the API and allows for remote calls.
You can read resources like thumbnails directly from the Reader without forcing file I/O.
We will be expanding on this model over time.


src/c2pa.udl
```diff
@@ -2,38 +2,103 @@
...
+ /// [Throws=Error]
+ /// string read_file([ByRef] string path, string? data_dir);
+ /// [Throws=Error]
+ /// string read_ingredient_file([ByRef] string path, [ByRef] string data_dir);
+ /// [Throws=Error]
+ /// sequence<u8> sign_file([ByRef] string source, [ByRef] string dest, [ByRef] string manifest, [ByRef] SignerInfo signer_info, string? data_dir);
+};
+
+[Error]
+interface Error {
```
dkozma, line 8: Do you know what happens if this list falls out of sync with c2pa-rs?

src/c2pa.udl
```diff
@@ -2,38 +2,103 @@
...
+  string from_stream([ByRef] string format, [ByRef] Stream reader);
+
+  [Throws=Error]
+  string json();
+
+  [Throws=Error]
+  u64 resource_to_stream([ByRef] string uri, [ByRef] Stream stream);
 };
 
+callback interface SignerCallback {
```
dkozma, line 67: It does seem like we might want a c2pa-uniffi or have this someplace where Swift and Kotlin can access it. This might be another vote for an eventual monorepo that @dyro was kicking around.

src/error.rs
```diff
@@ -0,0 +1,126 @@
...
+    Other { reason: String },
+    #[error("Remote {reason}")]
+    RemoteManifest { reason: String },
+    #[error("ResourceNotFound {reason}")]
+    ResourceNotFound { reason: String },
+    #[error("RwLock")]
+    RwLock,
+    #[error("Signature {reason}")]
+    Signature { reason: String },
+    #[error("Verify{reason}")]
```
dkozma: Nit
```suggestion
    #[error("Verify {reason}")]
```

src/error.rs
```diff
@@ -0,0 +1,126 @@
+use thiserror::Error;
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Error, Debug)]
+/// Defines all possible errors that can occur in this library
+pub enum Error {
```
dkozma, line 6: Is this also something that can be shared by the other uniffi-based bindings?

src/lib.rs
```diff
@@ -12,14 +12,179 @@
...
+mod error;
+pub use error::{Error, Result};
+#[cfg(feature = "v1")]
+mod json_api;
+#[cfg(feature = "v1")]
+pub use json_api::{read_file, read_ingredient_file, sign_file};
+#[cfg(feature = "v1")]
+mod signer_info;
+#[cfg(feature = "v1")]
+pub use signer_info::{CallbackSigner, SignerCallback, SignerConfig, SignerInfo};
```
dkozma, line 29: It looks like you are still keeping the V1 stuff in here? It might be good to put this under a `v1` namespace or something so it's clear to other people referencing the code (looking at people such as myself) which parts are code related to the new way.

dkozma (commented):
This looks great, however I'm just wondering how much of this can be shared with other uniffi-based bindings (Swift, Kotlin) and if we should put that in a sharable place now, since Swift will be coming in the near future.

src/c2pa.udl
gpeacock: It won't compile if they don't match up.

src/error.rs
gpeacock: Ideally the Rust API would use the same set  of errors as well. So I'd like to merge this into c2pa-rs and have these be the standard errors for all. Note that this needs that error review task have on the list.

src/lib.rs
gpeacock: I'll strip this out. I was keeping it in to make sure I'd covered the functional equivalents.

src/c2pa.udl
gpeacock: I agree, we should find a way to share the uniffi based bindings

dyro (approved):
The Rust and Uniffi code looks solid — I have little experience with Python, so will defer to others for that.

dkozma (approved):
LGTM as well - we can consolidate the uniffi code once this moves into the bindings repo.

PR #29, opened 2024-08-10
c2pa_api: remove double return statement

iameli-streams:
I'm assuming this was a typo?


gpeacock (approved):

iameli-streams:
Looks like this was fixed upstream.

PR #34, opened 2024-09-14
Performance improvement when signing using file paths.

gpeacock:
sign_file now uses Rust file i/o for performance.
sign_ps256 takes bytes instead of a file path.
added benchmark test for signing.


dyro (approved):
Nice use of benchmark 👍

PR #35, opened 2024-09-16
update perl to 5.30 in the build container

andrewhalle:
There's currently an [OpenSSL build failure][0] caused by the manylinux container having too old a version of Perl.

The workaround is to update to a newer version of Perl.

[0]: https://github.com/openssl/openssl/issues/25366


gpeacock (approved):

dyro:
Closing as we went with the openssl-src fix instead.

PR #39, opened 2024-09-18
Release notes for v0.5.2 

crandmck:
Release notes for v 0.5.2.

Also added a note about how to determine what version you've installed and how to update.


README.md
```diff
@@ -292,10 +302,22 @@ deactivate
 
 ## Release notes
 
+### Version 0.5.2
+
+New features:
+
+- DER fix
```
gpeacock: -  Allow EC signatures in DER format from signers and verify signature format during validation.
- Fixes bug in signing BMFF formats.
- Adds the ability to verify PDF files (but not to sign them).
- 2x or better speed improvement on sign_file when using file paths (uses native Rust file I/O).
- fixes for RIFF and GIF formats.

gpeacock (changes requested):

gpeacock (approved):

PR #43, opened 2024-10-01
feat: Adds Reader.from_manifest_data_and_stream along with set_remote_url and set_no_embed.

gpeacock:


c2pa/c2pa_api/c2pa_api.py
```diff
@@ -36,9 +36,12 @@
 # reader = Reader("image/jpeg", open("test.jpg", "rb"))
 # json = reader.json()
 class Reader(api.Reader):
-    def __init__(self, format, stream):
+    def __init__(self, format, stream, manifest_data=None):
         super().__init__()
-        self.from_stream(format, C2paStream(stream))
+        if manifest_data is not None:
```
tmathern, line 41: Is there any chance that the manifest_data could be blank (e.g. ` ` or any variation of whatever could be considered blank). Do we need special care around that?

src/lib.rs
```diff
@@ -121,6 +134,25 @@ impl Builder {
         Ok(())
     }
 
+    /// Set to true to disable embedding a manifest 
+    pub fn set_no_embed(&self) -> Result<()> {
+        if let Ok(mut builder) = self.builder.try_write() {
+            builder.set_no_embed(true);
+        } else {
+            return Err(Error::RwLock);
```
tmathern, line 142: Question: Is this to poison the lock or... ?

c2pa/c2pa_api/c2pa_api.py
gpeacock: This option is allow verifying a sidecar or remote manifest. If we have retrieved a .c2pa manifest store from our cloud or a sidecar, we can pass it in along with an associated asset to see if it verifies. If the manifest_data is not a correctly structured binary .c2pa we will return errors reporting it.
I should add a ticket somewhere to do more type checking at a higher level, If you pass the wrong type of parameter, it will get caught, but the error messages are obscure ones from the binding layer.

src/lib.rs
gpeacock: This is here because the uniffi tool we use to generate python bindings does not support mutable parameters. So we add an internal check to ensure there is only one mutable access at a time.

crandmck:
Should we add something to the docs (i.e. README) about these new functions?

src/lib.rs
dyro: We could use atomics here in `Buidler` to get around this eventually, if we end up having to have these binding layers hold on to a Mutex because of the Uniffi requirements of no mut vars as params. Not something that needs to be done for this PR, but it may make our lives easier down the road. 

```rust
use std::sync::atomic::{AtomicBool, Ordering};

struct Builder {
    no_embed: AtomicBool
}

impl Builder {
    pub fn set_no_embed(&self, value: bool) {
        self.no_embed.store(value, Ordering::Relaxed)
    }
}


fn main() {
    let b = Builder {
        no_embed: AtomicBool::new(true)
    };
    
    
    b.set_no_embed(true);
    
    // note how b, defined above, is not mutable. 
    std::thread::scope(|s| {
        let _ = s.spawn(|| b.set_no_embed(false));
        let _ = s.spawn(|| b.set_no_embed(true));
        let _ = s.spawn(|| b.set_no_embed(false));
    });
}
```

dyro (approved):
Looks good! Posted an idea now how to prevent the Mutex in bindings layers, but definitely not required to get this PR in.

scouten-adobe (approved):
No concerns on the Rust side. I don't know enough Python to review that side.

gpeacock:
> Looks good! Posted an idea now how to prevent the Mutex in bindings layers, but definitely not required to get this PR in.

Thanks for the suggestion. I'd rather not redefine the Builder structure for the Python bindings. But we could look into alternatives if this locking turns out to be an issue.

PR #45, opened 2024-10-07
Fix order of arguments in Reader init

barnjamin:
The arguments are swapped here based on what is generated for the bindings

Also wrapping the `stream` argument in a C2paStream like the other branch in the init fn


gpeacock (commented):
Thanks for catching this, I just pushed another PR with this fix along with unit test fixes that should have caught it.

gpeacock:
This fix was made along with others in https://github.com/contentauth/c2pa-python/pull/47
Thank you!

barnjamin:
Sure! I realize #47 is merged but I made a comment on it anyway https://github.com/contentauth/c2pa-python/pull/47#discussion_r1793286916

PR #47, opened 2024-10-09
fix: fixes bug in from_manifest_and_stream.

gpeacock:
Renames unit_tests so it will run with pytest.

### DIFF
--- a/c2pa/c2pa_api/c2pa_api.py
+++ b/c2pa/c2pa_api/c2pa_api.py
@@ -39,7 +39,7 @@ class Reader(api.Reader):
     def __init__(self, format, stream, manifest_data=None):
         super().__init__()
         if manifest_data is not None:
-            self.from_manifest_data_and_stream(format, manifest_data, stream)
+            self.from_manifest_data_and_stream(manifest_data, format, stream)
         else:
             self.from_stream(format, C2paStream(stream))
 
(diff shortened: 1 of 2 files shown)


tmathern (approved):

c2pa/c2pa_api/c2pa_api.py
```diff
@@ -39,7 +39,7 @@ class Reader(api.Reader):
     def __init__(self, format, stream, manifest_data=None):
         super().__init__()
         if manifest_data is not None:
-            self.from_manifest_data_and_stream(format, manifest_data, stream)
+            self.from_manifest_data_and_stream(manifest_data, format, stream)
```
barnjamin, line 42: Shouldn't the `stream` here also be wrapped in a `C2paStream`?

barnjamin: The unit test modified in this PR will fail with a hash mismatch if `validation_status` is checked:
```python
self.assertNotIn("validation_status", reader.json())
```

Wrapping the stream arg before calling init makes the test pass but it seems like it should be done in the `Reader` init
image

barnjamin: Filed a [PR](https://github.com/contentauth/c2pa-python/pull/48) for this so it didn't get lost

PR #54, opened 2024-10-23
Make underlying plumbing more generic

tmathern:
Make underlying things more generic, for a more flexible API.


c2pa/c2pa_api/c2pa_api.py
```diff
@@ -193,7 +192,36 @@ def __init__(self, callback):
...
+        case "SigningAlg.PS512":
+            return api.SigningAlg.PS512
+        case "SigningAlg.ED25519":
+            return api.SigningAlg.ED25519
+        case _:
+            raise ValueError("Unsupported signing algorithm: " + str(alg))
+
+def create_remote_signer(callback, alg=None):
+    alg_to_use = alg
+    if alg_to_use is not None:
```
tmathern: Would we even want that? The option to specify an algorithm? It's nice, but a little bit "use at your own risk"

src/callback_signer.rs
```diff
@@ -10,7 +10,8 @@
 // specific language governing permissions and limitations under
 // each license.
 
-use c2pa::SigningAlg;
+use c2pa::{SigningAlg, Signer};
+use log::debug;
```
tmathern, line 14: What can I say, I love my debug logs.

src/callback_signer.rs
```diff
@@ -24,7 +25,41 @@ pub trait SignerCallback: Send + Sync {
...
 /// Uniffi callbacks are only supported as a method in a structure, so this is a workaround
 pub struct CallbackSigner {
-    signer: c2pa::CallbackSigner,
+    signer: Box<dyn Signer + Sync + Send>,
+}
+
+pub struct RemoteSigner {
+    signer_callback : Box<dyn SignerCallback>,
+    alg: SigningAlg,
+    reserve_size: u32,
```
tmathern, line 34: reserve_size really should be a `usize`, but usize doesn't map well in uniffi (it's not a supported type). So I picked u32 so at least most of the time with type coercions we should not encounter truncation issues. But it introduces limitations (since it's technically a usize)...

src/lib.rs
```diff
@@ -119,6 +120,8 @@ impl Builder {
     ///
     /// Uniffi does not support constructors that return errors
     pub fn new() -> Self {
+        env_logger::init();
```
tmathern: Temporary, will remove - needed to initialize the logs starting somewhere...

Cargo.toml
```diff
@@ -19,6 +19,8 @@ serde_json = "1.0"
 thiserror = "1.0.49"
 uniffi = "0.24.1"
 openssl-src = "=300.3.1" # Required for openssl-sys
+log = "0.4.21"
+env_logger = "0.11.3"
```
tmathern, line 23: Will go when I remove temp log initialization

c2pa/c2pa_api/c2pa_api.py
```diff
@@ -193,7 +192,33 @@ def __init__(self, callback):
...
+        case "SigningAlg.PS384":
+            return api.SigningAlg.PS384
+        case "SigningAlg.PS512":
+            return api.SigningAlg.PS512
+        case "SigningAlg.ED25519":
+            return api.SigningAlg.ED25519
+        case _:
+            raise ValueError("Unsupported signing algorithm: " + str(alg))
+
+def create_remote_signer(callback):
```
gpeacock, line 220: We need to add documentation for this method. It may not need an example, but something like:
`Creates a special case signer that uses direct COSE handling.`

gpeacock (dismissed):
Looks good but needs a some extra comments on the new functionality.

c2pa/c2pa_api/c2pa_api.py
tmathern: Got it - I added the doc.

gpeacock: Actually i just meant you needed a comment in the code here. But updating the Readme is a good idea too!

gpeacock (approved):

PR #56, opened 2024-10-28
chore: adds tests, clippy, formatting, and auditing to CI

dyro:


Cargo.toml
```diff
@@ -20,7 +22,6 @@ thiserror = "1.0.49"
 uniffi = "0.24.1"
 openssl-src = "=300.3.1" # Required for openssl-sys
 log = "0.4.21"
-env_logger = "0.11.3"
```
dyro, line 23: Unused dependency.

Cargo.toml
```diff
@@ -9,6 +9,8 @@ authors = ["Gavin Peacock <gpeacock@adobe.com"]
 name = "c2pa"
 crate-type = ["cdylib"]
 
+[package.metadata.cargo-udeps.ignore]
+normal = ["openssl-src"]
```
dyro, line 13: We use this to explicitly set the openssl-src dependency.

src/json_api.rs
```diff
@@ -1,115 +0,0 @@
...
-        }
-        let result = read_file(&path, Some(data_dir.to_owned()));
-        assert!(result.is_ok());
-        let json_report = result.unwrap();
-        println!("{}", json_report);
-        assert!(json_report.contains("C.jpg"));
-        assert!(PathBuf::from(data_dir).exists());
-        assert!(json_report.contains("thumbnail"));
-    }
-}
```
dyro, line 115: Removes this because it wasn't being used in the new APIs, and we've removed the v1 and add_thumbnails features this was reliant upon for being included.

src/lib.rs
```diff
@@ -19,14 +19,6 @@ pub use c2pa::SigningAlg;
...
 mod error;
 pub use error::{Error, Result};
-#[cfg(feature = "v1")]
-mod json_api;
-#[cfg(feature = "v1")]
-pub use json_api::{read_file, read_ingredient_file, sign_file};
-#[cfg(feature = "v1")]
-mod signer_info;
-#[cfg(feature = "v1")]
-pub use signer_info::{CallbackSigner, SignerCallback, SignerConfig, SignerInfo};
```
dyro, line 29: This features had already been removed from the crate. Removing the code so clippy is happy :)

src/lib.rs
```diff
@@ -57,6 +49,12 @@ pub struct Reader {
     reader: RwLock<c2pa::Reader>,
 }
 
+impl Default for Reader {
```
dyro, line 52: Clippy suggestion

src/signer_info.rs
```diff
@@ -1,60 +0,0 @@
...
-    pub fn signer(&self) -> Result<Box<dyn Signer>> {
-        create_signer::from_keys(
-            &self.sign_cert,
-            &self.private_key,
-            self.alg()?,
-            self.ta_url.clone(),
-        )
-        .map_err(Error::from_c2pa_error)
-    }
-}
```
dyro, line 60: Removing for the same reason as removing the json_api. This wasn't being used and it was hidden behind a feature flag that was removed. Lemme know if I should add this back, @gpeacock

src/streams.rs
```diff
@@ -70,7 +70,7 @@ impl<'a> StreamAdapter<'a> {
 impl<'a> From<&'a dyn Stream> for StreamAdapter<'a> {
     #[allow(invalid_reference_casting)]
     fn from(stream: &'a dyn Stream) -> Self {
-        let stream = &*stream as *const dyn Stream as *mut dyn Stream;
+        let stream = stream as *const dyn Stream as *mut dyn Stream;
```
dyro, line 73: The old version and new version are the same type, as pointed out by clippy. Lemme know if I'm missing something, @gpeacock. I know the binding stuff can be crazy. Maybe the `&*` is required anyway?

src/error.rs
```diff
@@ -96,8 +96,6 @@ impl Error {
             ClaimVerification(_) | InvalidClaim(_) | JumbfParseError(_) => {
                 Self::Verify { reason: err_str }
             }
-            #[cfg(feature = "add_thumbnails")]
-            ImageError => Self::ImageError(err_str),
```
dyro, line 100: Same here: this feature was removed when we swapped to the V24 API 👍

tmathern (approved):

src/streams.rs
gpeacock: I'll trust clippy on this one. That's a remnant of older code.

src/lib.rs
gpeacock: Yes, it should be fine to remove this now. No one seems overly bothered by it going away.

gpeacock (approved):

PR #59, opened 2024-11-05
fix: Export CallbackSigner

gpeacock:
fix version based unit tests
Add a title from the filename if no title exists in Builder add_ingredient_file()

### DIFF
--- a/c2pa/c2pa_api/c2pa_api.py
+++ b/c2pa/c2pa_api/c2pa_api.py
@@ -114,6 +114,10 @@ class Builder(api.Builder):
 
     def add_ingredient_file(self, ingredient, path):
         format = os.path.splitext(path)[1][1:]
+        if "title" not in ingredient:
+            if isinstance(ingredient, str):
+                ingredient = json.loads(ingredient)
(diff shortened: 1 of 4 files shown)


c2pa/c2pa_api/c2pa_api.py
```diff
@@ -114,6 +114,10 @@ def add_ingredient(self, ingredient, format, stream):
 
     def add_ingredient_file(self, ingredient, path):
         format = os.path.splitext(path)[1][1:]
+        if "title" not in ingredient:
+            if isinstance(ingredient, str):
+                ingredient = json.loads(ingredient)
```
tmathern, line 119: Should this be wrapped in try/catch in case the json loading fails?

tmathern (approved):

c2pa/c2pa_api/c2pa_api.py
gpeacock: Possibly, but caller should be catching exceptions like this. I'll do an updated pass on all the errors and exceptions when I review the main SDK errors.

PR #65, opened 2024-11-14
chore: Update to latest maturin/uniffi

tmathern:
Update to latest maturin+uniffi. 

Follow-up: https://github.com/contentauth/c2pa-python/pull/66


Makefile
```diff
@@ -0,0 +1,9 @@
+# For Python bindings ===========================================================
```
gpeacock, line 1: Very nice to have this. It adds support for make to the dependencies, but I suppose it could be seen as more of a convenience. (On windows you need to install make support)

gpeacock (approved):
I've been meaning to update for a long time, nice to get this done!

PR #74, opened 2024-12-03
feat: Adds load_settings and load_settings_file functions

gpeacock:
Includes a sample loading a trust list.


src/lib.rs
```diff
@@ -102,6 +107,14 @@ impl Reader {
         }
     }
 
+    // pub fn validation_state(&self) -> Result<ValidationState> {
```
tmathern: Debug leftover?

tests/fixtures/settings.toml
```diff
@@ -0,0 +1,229 @@
+# This sample c2pa settings file enables a trust list.
```
tmathern, line 1: I wonder if we should explain why TOML was picked as a format (I remember there were some challenges because of the type of embedded data. We may want to spare the gotcha to others?)

tmathern (approved):

tmathern (approved):

tmathern (approved):

PR #75, opened 2024-12-03
chore: Run formatter on all the python files

gpeacock:
Also removes file close warnings in tests.


c2pa/__init__.py
```diff
@@ -11,8 +11,35 @@
...
+from .c2pa_api import (
+    Reader,
+    Builder,
+    create_signer,
+    create_remote_signer,
+    sign_ps256,
+    load_settings_file,
+)
+from .c2pa import Error, SigningAlg, CallbackSigner, sdk_version, version, load_settings
+from .c2pa.c2pa import (
```
tmathern, line 23: I never saw the syntax with parenthesis before, hope it doesn't have adverse strange side-effects. After researching it, that's allowed only for Python 3.10 and later, so Python-wise should be fine.

src/lib.rs
```diff
@@ -102,6 +107,14 @@ impl Reader {
         }
     }
 
+    // pub fn validation_state(&self) -> Result<ValidationState> {
```
tmathern, line 110: Is it debug code?

tmathern (approved):
Not going to argue with a formatter

tmathern:
Closing as this is now obsolete.

PR #79, opened 2025-01-07
chore: updates c2pa version

dyro:


tmathern (approved):

dyro:
Going to merge even though the `udeps` check is failing. A team member is working on a fix that I'll integrate here when it's ready.

PR #83, opened 2025-01-22
chore: Update c2pa-rs version to latest

tmathern:
- Update c2pa-rs version to latest
- Update Python/s crypto version while at it
- Add missing test deps in requirements.txt

PR #84, opened 2025-01-22
chore: update c2pa-rs

tmathern:
Follow-up of https://github.com/contentauth/c2pa-python/pull/83/, with new c2pa-rs version.
This version of c2pa-rs allows us to drop the c2pa-crypto dependency.

### DIFF
--- a/src/callback_signer.rs
+++ b/src/callback_signer.rs
@@ -11,8 +11,6 @@
 // each license.
 
 use c2pa::{Signer, SigningAlg};
-// RawSigner is currently used only fully qualified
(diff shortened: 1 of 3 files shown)


src/callback_signer.rs
```diff
@@ -36,41 +34,6 @@ pub struct RemoteSigner {
     reserve_size: u32,
 }
 
-impl TimeStampProvider for RemoteSigner {}
```
gpeacock, line 39: oh so much better!

gpeacock (approved):

PR #91, opened 2025-02-21
docs: Update release notes with breaking change, other doc updates

crandmck:
Per #90 
- There was a release: v0.6.1 in Nov. https://pypi.org/project/c2pa-python/, but it doesn't appear in https://github.com/contentauth/c2pa-python/releases ... why?
    - For consistency, shouldn't https://github.com/contentauth/c2pa-python/releases match with https://pypi.org/project/c2pa-python/#history ?
- Is there more we can put in these release notes?  There isn't much there!

Also we need to revise https://github.com/contentauth/c2pa-python/blob/main/docs/usage.md#signing-function because the `sign_ps256` function [has moved](https://github.com/contentauth/c2pa-python/blob/main/c2pa/c2pa_api/c2pa_api.py#L244).  Is the general info still accurate?  Gavin said something about a remote signer or callback signer.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -6,6 +6,10 @@
 
 See [Release tag 0.6.0](https://github.com/contentauth/c2pa-python/releases/tag/v0.6.0).
 
+### Breaking changes
+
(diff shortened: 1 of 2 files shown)


tmathern:
As discussed: v0.6.1 => there was a release, we forgot to tag it.

PR #94, opened 2025-03-27
chore: repair publishing process

gpeacock:

PR #100, opened 2025-05-07
ci: Update setup for publishing

tmathern:
Replaces https://github.com/contentauth/c2pa-python/pull/94.

Details about the update are logged in the comments on the PR (see below).

Example runs:
- Demo 1: (debug) https://github.com/contentauth/c2pa-python/actions/runs/14895338961 (skips real upload, since I did not bump the version number yet and it exists in PyPI. But we get passed the point when an upload would have happened).
- Demo 2: (without debug logs) https://github.com/contentauth/c2pa-python/actions/runs/14895530386 (set to skip-existing files, so fails after attempting upload, as expected)
- Demo 3: (without debug logs) https://github.com/contentauth/c2pa-python/actions/runs/14895714503 (skips real upload, since I did not bump the version number yet and it exists in PyPI. Except that, showcases a flow running).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Build
+name: build
 
 on:
   pull_request:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -381,4 +381,6 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           packages_dir: dist
+          username: __token__
+          password: ${{ secrets.PYPI_TOKEN }}
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -369,10 +369,16 @@ jobs:
       # IMPORTANT: this permission is mandatory for Trusted Publishing
       id-token: write
     steps:
+      - uses: actions/checkout@v4
       - uses: actions/download-artifact@v4
         with:
           pattern: wheels-*
-          name: wheels
-      - name: Publish to PyPI
+          path: dist
+          merge-multiple: true
+      - name: List contents of dist directory
+        run: ls -la dist/
+      - name: Publish to PyPi
         uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages_dir: dist
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -384,6 +384,10 @@ jobs:
           echo "Event name: ${{ github.event_name }}"
       - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
+        env:
+          TWINE_VERBOSE: 1
+          TWINE_DEBUG: 1
+          PYTHONUNBUFFERED: 1
         with:
           packages-dir: dist
           verbose: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -359,14 +359,11 @@ jobs:
 
   release:
     name: Release
-
     if: startsWith(github.ref, 'refs/tags/') || github.event_name == 'workflow_dispatch' && github.event.inputs.publish == 'true'
-
     runs-on: ubuntu-latest
     environment: pypipublish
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -1,4 +1,4 @@
-name: Build
+name: build
```
tmathern: Humor me while I debug publishing. Will restore name with uppercase if it's not causing issues.

.github/workflows/upload-test.yml
```diff
@@ -0,0 +1,98 @@
+name: Upload Test for Pypi.
```
tmathern: Note: Came in when I merged in latest main, did not make changes here in that PR.

Makefile
```diff
@@ -15,4 +15,8 @@ build-python:
 
 test:
 	python3 ./tests/test_unit_tests.py
-	python3 ./tests/test_api.py
\ No newline at end of file
+	python3 ./tests/test_api.py
+
+publish: release
```
tmathern: Came in with latest main too

pyproject.toml
```diff
@@ -3,7 +3,7 @@ requires = ["maturin>=1.7.4,<2.0","uniffi_bindgen>=0.28,<0.30"]
 build-backend = "maturin"
 
 [project]
-name = "c2pa"
+name = "c2pa-python"
```
tmathern, line 6: That name needs to match what is in PyPI for Trusted publishing.

pyproject.toml
```diff
@@ -27,4 +27,7 @@ urls = {homepage = "https://contentauthenticity.org", repository = "https://gith
 [project.optional-dependencies]
 test = [
   "pytest < 5.0.0"
-]
\ No newline at end of file
+]
+
+[tool.maturin]
+module-name = "c2pa"
```
tmathern, line 33: Because we had to update the name above for trusted publishing, now we need this to "fix" the package name.

.github/workflows/build.yml
tmathern: Will need to restore the name to a nicer format

.github/workflows/build.yml
```diff
@@ -359,15 +359,13 @@ jobs:
 
   release:
     name: Release
-
     if: startsWith(github.ref, 'refs/tags/') || github.event_name == 'workflow_dispatch' && github.event.inputs.publish == 'true'
-
     runs-on: ubuntu-latest
-    environment: Publish
+    environment: pypipublish
```
tmathern, line 364: The environment needs to match what is set in trusted publishing on PyPI (and it's case-sensitive). Unfortunately, it can do *only* lower case environments, so the env in github needs to be all lowercase too.

.github/workflows/build.yml
```diff
@@ -359,15 +359,13 @@ jobs:
...
     needs: [linux, windows, macos_x86, macos_aarch64, sdist]
     permissions:
-      # IMPORTANT: this permission is mandatory for Trusted Publishing
       id-token: write
+      contents: read
```
tmathern, line 368: Added one more permission so we're sure the token can be used, and the contents we're dealing with, read

.github/workflows/build.yml
gpeacock: I'm really glad you found this!

.github/workflows/build.yml
```diff
@@ -359,21 +359,28 @@ jobs:
...
     steps:
+      - uses: actions/checkout@v4
       - uses: actions/download-artifact@v4
         with:
           pattern: wheels-*
-          name: wheels
+          path: dist
+          merge-multiple: true
+      - name: List contents of dist directory
+        run: ls -la dist/
```
gpeacock, line 377: probably don't need this now that it works, but I think I'd leave it in anyway, just to be sure

gpeacock (approved):
Thank you so much for this. It is LONG overdue!

.github/workflows/build.yml
tmathern: Sounds good. We can remove it in subsequent PRs. And you're right, it's nice to have here to see what goes in.

PR #102, opened 2025-05-13
chore: Allow new license

tmathern:
Allow new license


gpeacock (approved):
Did we get the ok on this? If so then great!

PR #104, opened 2025-05-14
feat: Rewrite this repo to use pre-built c2pa_rs libraries and no dir…

gpeacock:
…ect Rust compilation


pyproject.toml
```diff
@@ -1,33 +1,37 @@
...
-dependencies = ["cffi"]
-requires-python = ">=3.7"
+version = "0.10.15"
+requires-python = ">=3.10"
 description = "Python bindings for the C2PA Content Authenticity Initiative (CAI) library"
 readme = { file = "README.md", content-type = "text/markdown" }
-license = {file = "LICENSE-MIT"}
-keywords = ["C2PA", "CAI", "content credentials", "metadata", "provenance"]
+license = { text = "MIT OR Apache-2.0" }
+authors = [{ name = "Gavin Peacock", email = "gvnpeacock@adobe.com" }]
```
gpeacock: You can add your name here if you want

tests/test_unit_tests_threaded.py
```diff
@@ -0,0 +1,2239 @@
...
+            "claim_generator": "python_test_thread1",
+            "claim_generator_info": [{
+                "name": "python_test_1",
+                "version": "0.0.1",
+            }],
+            "format": "image/jpeg",
+            "title": "Python Test Image 1",
+            "ingredients": [],
+            "assertions": [
+                {'label': 'stds.schema-org.CreativeWork',
```
gpeacock: we should move away from CreativeWork as the default test example and also ensure we always add a c2pa.created or c2pa.opened, assertion.

tests/test_unit_tests_threaded.py
```diff
@@ -0,0 +1,2239 @@
+# Copyright 2025 Adobe. All rights reserved.
```
gpeacock, line 1: I don't know if anyone has asked or expected us to work with multithreaded workflows, but it's nice to know we do. My main concern is that we will need to maintain all the test cases we add. But as long as we have the resources to do it, I see no harm.

tests/test_unit_tests.py
```diff
@@ -16,122 +16,897 @@
...
-        self.assertIn("0.9.0", sdk_version())
+    def test_sdk_version(self):
+        self.assertIn("0.55.0", sdk_version())
 
 
 class TestReader(unittest.TestCase):
+    def setUp(self):
+        # Use the fixtures_dir fixture to set up paths
+        self.data_dir = os.path.join(os.path.dirname(__file__), "fixtures")
+        self.testPath = os.path.join(self.data_dir, "C.jpg")
```
gpeacock: We should set up some kind of constant to refer to the C.jpg filename here since we reference it may other places in the tests.

tests/test_unit_tests.py
```diff
@@ -16,122 +16,897 @@
...
+        reader.close()
+        with self.assertRaises(Error):
+            reader.resource_to_stream("", io.BytesIO(bytearray()))
+
+    def test_read_all_files(self):
+        """Test reading C2PA metadata from all files in the fixtures/files-for-reading-tests directory"""
+        reading_dir = os.path.join(self.data_dir, "files-for-reading-tests")
+
+        # Map of file extensions to MIME types
+        mime_types = {
```
gpeacock, line 120: Fyi, the apis that accept a format work with either an extension or a mime type. We may to export something like this conversion in the toolkit.

tests/test_unit_tests.py
```diff
@@ -16,122 +16,897 @@
...
+        # Create a local Ps256 signer with certs and a timestamp server
+        self.signer_info = C2paSignerInfo(
+            alg=b"es256",
+            sign_cert=self.certs,
+            private_key=self.key,
+            ta_url=b"http://timestamp.digicert.com"
+        )
+        self.signer = Signer.from_info(self.signer_info)
+
+        self.testPath = os.path.join(self.data_dir, "C.jpg")
```
gpeacock: Same comment about using constants for the filenames. If we are abstracting the setup we should also abstract the filenames we test for.

tests/test_unit_tests_threaded.py
tmathern: Ack - I update to use a "custom" one for which I can control the data to verify possible scrambling. And I hope I made it clear enough it's custom for the thread tests (and not part of the spec), to specifically test for scrambling.

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,150 @@
...
+          Write-Host "Artifacts directory contents:"
+          Get-ChildItem -Recurse -Path artifacts
+          Write-Host "src/c2pa/libs directory contents:"
+          Get-ChildItem -Recurse -Path src/c2pa/libs
+
+          # Build wheel
+          python setup.py bdist_wheel --plat-name win_amd64
+
+      - name: Build macOS wheel (Apple Silicon)
+        if: runner.os == 'macOS' && runner.arch == 'arm64'
```
gpeacock, line 113: I think we will need to support x86 macos, but it can probably wait for now.

tests/benchmark.py
```diff
@@ -1,9 +1,30 @@
...
+import json
+import pytest
+from c2pa import Reader, Builder, Signer, C2paSigningAlg, C2paSignerInfo
+
 PROJECT_PATH = os.getcwd()
 
-testPath = os.path.join(PROJECT_PATH, "tests", "fixtures", "C.jpg")
+# Test paths
+test_path = os.path.join(PROJECT_PATH, "tests", "fixtures", "C.jpg")
+output_path = "target/python_out.jpg"
```
gpeacock: the target folder was a Rust thing, I think we should just use a temp folder for tests here.

.github/workflows/build-wheel.yml
tmathern: Can x86 support using the universal libs? I think so? That build target may be easier to get our hands on.

tests/benchmark.py
tmathern: Oh `target` is handled like a temp folder. I clean it up in the end. I cna move it into the test folder.

tests/test_unit_tests.py
tmathern: (Done, will be in next pushed changes).

tests/test_unit_tests.py
tmathern: Ack, added the test to showcase it (I want the tests to be examples of our capabilities).

tests/benchmark.py
tmathern: Renamed & moved it to be a folder called temp under the tests folder.

PR #105, opened 2025-05-14
feat: Update c2pa.py

tmathern:
- Update c2pa.py with latest changes from the POC branch
- Update makefile
- Clean up tests so resources are properly released


src/c2pa/lib.py
```diff
@@ -0,0 +1,103 @@
...
+    ]
+
+    if lib_name:
+        # If specific library name is provided, only load that one
+        lib = _load_single_library(lib_name, possible_paths)
+        if not lib:
+            raise RuntimeError(f"Could not find {lib_name} in any of the search paths")
+        return lib
+
+    # Default paht (no library name provided in the environment)
```
gpeacock: Default path (typo)

src/c2pa/lib.py
```diff
@@ -0,0 +1,103 @@
...
+            # Continue with normal loading if environment variable library name fails
+
+    # Try to find the libraries in various locations
+    possible_paths = [
+        # Current directory
+        Path.cwd(),
+        # Package directory
+        Path(__file__).parent,
+        # Additional library directory
+        Path(__file__).parent / "lib",
```
gpeacock: When I create a wheel, the libraries are put under a libs directory in the src folder, so we either need to add that path here or change the libs reference to lib so that this will work in that environment.

gpeacock (changes requested):

src/c2pa/lib.py
tmathern: Updated!

scripts/download_artifacts.py
```diff
@@ -5,12 +5,14 @@
...
 import zipfile
 import io
+import shutil
 
 # Constants
 REPO_OWNER = "contentauth"
 REPO_NAME = "c2pa-rs"
 GITHUB_API_BASE = "https://api.github.com"
-ARTIFACTS_DIR = Path("artifacts")
+SCRIPTS_ARTIFACTS_DIR = Path("scripts/artifacts")
```
tmathern, line 16: So we can run this from the root of the repo. I used this instead of `build.py` because `build.py` seems broken, whereas that one works. Let me know if `build.py` was meant to stay or not - but then it likely needs debugging.

gpeacock: I did work on build.py to get it functional, but I can merge that in.

src/c2pa/lib.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Returns:
+        The loaded library or None if loading failed
+    """
+    if sys.platform == "darwin":
+        c2pa_lib_name = "libc2pa_c.dylib"
+    elif sys.platform == "linux":
+        c2pa_lib_name = "libc2pa_c.so"
+    elif sys.platform == "win32":
+        c2pa_lib_name = "c2pa_c.dll"
```
gpeacock, line 203: I'm never sure on windows if this is .dll or .dll.lib, but we can sort that out later.

src/c2pa/c2pa.py
```diff
@@ -436,55 +520,207 @@ def sign_file(
...
         self._file = file
-        
+        self._stream = None  # Initialize to None to track if stream was created
+        self._closed = False  # Track if the stream has been closed
+        self._initialized = False  # Track if stream was successfully initialized
+
+        # print(f'## Created stream {self._stream_id} for file {self._file}')
+
+        # Pre-allocate error message strings to avoid string formatting overhead
+        self._error_messages = {
```
gpeacock, line 571: I'm not sure why we would be concerned about the performance of error message generation, but I guess it does no particular harm

src/c2pa/c2pa.py
```diff
@@ -494,203 +730,371 @@ def flush_callback(ctx):
...
+        Raises:
+            C2paError: If there was an error creating the reader
+            C2paError.Encoding: If any of the string inputs contain invalid UTF-8 characters
         """
+
         self._reader = None
         self._own_stream = None
-        self._strings = []  # Keep encoded strings alive
+        self._error_messages = {
+            'unsupported': "Unsupported format",
```
gpeacock, line 825: We probably want to line these up with the same format of the messages we return in the C++ api. I'm not sure which is better, but I think I use CamelCase for the identifiers in C++

src/c2pa/c2pa.py
```diff
@@ -494,203 +730,371 @@ def flush_callback(ctx):
...
+
+    def __init__(self, signer_ptr: ctypes.POINTER(C2paSigner)):
+        """Initialize a new Signer instance.
+
+        Note: This constructor is not meant to be called directly.
+        Use from_info() or from_callback() instead.
+        """
+        self._signer = signer_ptr
+        self._closed = False
+        self._error_messages = {
```
gpeacock, line 1058: We need to huddle on error returns. I know there will be some differences depending on languages, but we also need to have consistency so that we don't have a completely different set of documentation for each binding. The errors are part of the API.

src/c2pa/c2pa.py
```diff
@@ -436,55 +520,207 @@ def sign_file(
     )
     return _handle_string_result(result)
 
-# Helper class for stream operations
 class Stream:
-    """High-level wrapper for C2paStream operations."""
+    # Class-level counter for generating unique stream IDs
+    # (useful for tracing streams usage in debug)
+    _next_stream_id = 0
```
gpeacock, line 526: These steam ids are just a debugging aid?

src/c2pa/c2pa.py
```diff
@@ -257,62 +334,68 @@ class Verify(Exception):
...
+    def __init__(self):
+        """Initialize an empty string container."""
+        pass
 
 def _handle_string_result(result: ctypes.c_void_p, check_error: bool = True) -> Optional[str]:
     """Helper function to handle string results from C2PA functions."""
     if not result:  # NULL pointer
         if check_error:
-            _handle_c2pa_error()
+            error = _lib.c2pa_error()
```
gpeacock, line 354: This feels odd to me. handle_string_result is designed specifically to return string results on methods that return strings. Using it everywhere when you already know you want to return an error is confusing. why not just call handle_c2pa_error if that is what you know you are doing?

gpeacock (commented):
There's a lot here. I added some feedback. Let me know what you think.

src/c2pa/lib.py
tmathern: I believe it _should_ be dll?

src/c2pa/c2pa.py
tmathern: Oh it's to avoid repeat allocation in memory. Very small tweak.

src/c2pa/c2pa.py
tmathern: Yes! Small but powerful to debug. It's to trace which stream is opened/closed/<put an action here> when. Helps to trace what is happening especially when the context managers (with statements) get involved and possibly nested.

src/c2pa/c2pa.py
tmathern: There is no `handle_c2pa_error` in that branch... ? ~Maybe in your branch... (I'll check and update if I see it there (not found?))?~ And I keep having to check `_lib.c2pa_error()` to not miss any error (remember we had the case where an error string would be reported there, but raise no error, or the called function's result would report being ok...). This is the additional layer to not miss any error with the current setup.

tmathern: I renamed `_parse_operation_result_for_error` but I am not totally sure we both aim for the same thing here. I aim to throw errors when errors happened (what _parse_operation_result_for_error now does), but you want to return strings? Returning strings may be an unexpected way to handle errors for people in Python more accustomed to dealing with exceptions being thrown.

src/c2pa/c2pa.py
tmathern: Ack - albeit those are specific to the Stream implementation on Python here.

src/c2pa/c2pa.py
tmathern: Ack, updating... (will be in next push).

src/c2pa/c2pa.py
gpeacock: I definitely want to throw errors when they happen! I created a bunch of odd little macros to handle things reliably in  the Rust C binding logic. They are oddly named and I need to document them and explain how they work. I've tried to follow old school C error handling in the C APIs, but it needs some work. Ideally every function would return an indication if there is an error or not, either by returning a negative number or returning null. If there's an error, we would always read the last error message and then throw a constructed error in the higher level language. The error string returned by c2pa_error is formatted to included an error name followed by a colon. The rest of the string contains the error message. The error name can be converted into a specific kind of C2paError in languages like Python. 
I know there some inconsistencies and I've been trying to eliminate them. I want all possible null parameters checked and a consistent policy on setting the last error so that we never end up with an empty error state.

tmathern: Ok. In that case, I would suggest to merge that PR into your branch, and rework from there (I really don't want the 2 branches to diverge too much).

src/c2pa/lib.py
tmathern: (dll.lib are for during the linking if I remember right)

PR #108, opened 2025-05-29
feat: Prepare publishing flow to PyPi, increase test coverage

tmathern:
- Publish wheels on PyPi(Test) matching what is supported in https://github.com/contentauth/c2pa-rs/releases/tag/c2pa-v0.55.0 (exception: macos intel excluded, as the runner is a tad old now). 
- Note that the workflow is written so it could be extended once we support more platforms (already set up with matrices to go through).
- Add unit test coverage (+fix some bugs found while writing the tests). Also updated the benchmark tests.
- Update examples to match new APIs.

_About the amount of files: I added many test fixtures that are tested "en masse" in the unit tests. It might be easier to review by ignoring those files while reading through!_


src/c2pa/lib.py
```diff
@@ -24,11 +24,13 @@
 )
```
tmathern, line 24: Autopep8 has formatting opinions (no functioncal code change).

src/c2pa/c2pa.py
```diff
@@ -43,6 +43,7 @@
     'c2pa_signature_free',
```
tmathern, line 44: I ran formating and Autopep8 has formatting opinions (no functioncal code change).

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,138 @@
...
+          fi
+
+          # Build wheel in Docker container
+          docker run --rm -v $PWD:/io $DOCKER_IMAGE bash -c "
+            yum install -y gcc gcc-c++ make &&
+            mkdir -p /io/artifacts /io/src/c2pa/libs &&
+            rm -rf /io/dist /io/build &&
+            cd /io &&
+            /opt/python/cp310-cp310/bin/pip install -r requirements.txt -r requirements-dev.txt &&
+            /opt/python/cp310-cp310/bin/pip install toml &&
```
ok-nick, line 65: I believe `toml` is already installed in the `requirements-dev.txt`?

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,138 @@
...
+            C2PA_LIBS_PLATFORM=\"${{ inputs.architecture == 'aarch64' && 'aarch64-unknown-linux-gnu' || 'x86_64-unknown-linux-gnu' }}\" /opt/python/cp310-cp310/bin/python scripts/download_artifacts.py c2pa-v0.55.0 &&
```
ok-nick: We may want to extract out the `v0.55.0` into a variable so it's clear how to update in the future. I think it would be best if it's extracted from the `Cargo.toml`, which is actually currently defined as `v0.51.0`

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,138 @@
...
+          mkdir -p src/c2pa/libs
+          rm -rf dist build
+
+          # Install dependencies
+          pip install -r requirements.txt
+          pip install -r requirements-dev.txt
+          pip install wheel
+
+          # Download native artifacts
+          python scripts/download_artifacts.py c2pa-v0.55.0
```
ok-nick: Same here for the version

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,138 @@
...
+          if (Test-Path dist) { Remove-Item -Recurse -Force dist }
+          if (Test-Path build) { Remove-Item -Recurse -Force build }
+
+          # Install dependencies
+          pip install -r requirements.txt
+          pip install -r requirements-dev.txt
+          pip install wheel
+
+          # Download native artifacts
+          python scripts/download_artifacts.py c2pa-v0.55.0
```
ok-nick: Same here for the version

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,138 @@
...
+            rm -f /io/dist/*-linux_*.whl
+          "
+
+          # Verify the wheel was built
+          echo "Contents of dist directory:"
+          ls -la dist/
+          echo "Number of wheels found:"
+          find dist -name "*.whl" | wc -l
+          echo "Wheel filenames:"
+          find dist -name "*.whl" -exec basename {} \;
```
ok-nick, line 81: Should this replace the "Log wheel filename" step since they have similar behavior?

.github/workflows/build.yml
```diff
@@ -48,73 +51,211 @@ jobs:
...
+          mkdir -p src/c2pa/libs
+          rm -rf dist/* build/*
+
+      - name: Download native artifacts
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Ensure the token is being used
+          echo "Using GitHub token for authentication"
+          python3 scripts/download_artifacts.py c2pa-v0.55.0
```
ok-nick: Here too

pyproject.toml
```diff
@@ -20,18 +20,23 @@ maintainers = [
   {name = "Gavin Peacock", email = "gpeacock@adobe.com"}
 ]
 urls = {homepage = "https://contentauthenticity.org", repository = "https://github.com/contentauth/c2pa-python"}
+dependencies = [
+    "wheel>=0.41.2",
+    "setuptools>=68.0.0",
+    "toml>=0.10.2",
+    "pytest>=7.4.0",
```
ok-nick, line 27: Did you mean to include pytest as a dependency and optional dependency?

pyproject.toml
```diff
@@ -20,18 +20,23 @@ maintainers = [
...
+    "cryptography>=41.0.0",
+    "requests>=2.0.0"
+]
```
ok-nick: Also, `pip` is able to install directly from a `pyproject.toml`, so I'm not sure if it's necessary to keep around the `requirements.txt` as well

.github/workflows/build-wheel.yml
tmathern: It is, but somehow not found. This won't hurt.

.github/workflows/build-wheel.yml
tmathern: Yes, but later.

.github/workflows/build-wheel.yml
tmathern: Later.

.github/workflows/build-wheel.yml
tmathern: Later.

.github/workflows/build-wheel.yml
tmathern: No - let it for debug at this step too. I needed multiple logs, and I have a hunch this will help again in the future.

.github/workflows/build.yml
tmathern: Later too.

pyproject.toml
tmathern: Good catch, updated.

pyproject.toml
tmathern: This is. We need it for deployment flows. Some are more limited/use the requirements.txt file. Also I want to leave both options open to users, even if for us it's slightly more maintenance.

ok-nick (approved):

.github/workflows/build-wheel.yml
tmathern: As promised: https://github.com/contentauth/c2pa-python/pull/113

.github/workflows/build-wheel.yml
tmathern: As promised: https://github.com/contentauth/c2pa-python/pull/113

.github/workflows/build-wheel.yml
tmathern: As promised: https://github.com/contentauth/c2pa-python/pull/113

.github/workflows/build.yml
tmathern: As promised: https://github.com/contentauth/c2pa-python/pull/113

pyproject.toml
tmathern: (Did some clean up though!)

README.md
```diff
@@ -12,26 +12,29 @@ This project provides a Python API for working with [C2PA](https://c2pa.org/) (C
...
-    │   ├── macosx_x86_64/ # macOS x64 libraries
-    │   ├── macosx_arm64/  # macOS ARM64 libraries
-    │   ├── linux_x86_64/  # Linux x64 libraries
-    │   └── linux_aarch64/ # Linux ARM64 libraries
-    ├── requirements.txt # Python dependencies
-    └── README.md       # Project documentation
+.
+├── .github/           # GitHub configuration files
+├── artifacts/         # Platform-specific libraries for building (per subfolder)
+│   └── linux_x86_64/ # Linux ARM64 libraries
```
gpeacock: I think this could be any of the platforms, but this should give the general idea

tmathern: Fix incoming in next push! (+further docs improvements)

PR #113, opened 2025-06-02
feat: Simplify version number handling

tmathern:
- Keep a text file with the c2pa-rs native version pinned
- Simplify version number trickle-through in the workflows


.github/workflows/build.yml
```diff
@@ -231,12 +245,13 @@ jobs:
...
     name: Build Windows wheel
     uses: ./.github/workflows/build-wheel.yml
-    needs: tests-windows
+    needs: [tests-windows, read-version]
     with:
       python-version: "3.10"
       architecture: ${{ matrix.target }}
       artifact-name: wheels-windows-${{ matrix.target }}
       runs-on: windows-latest
+      c2pa-version: "c2pa-v0.55.0"
```
ok-nick: Should this be the same as macOS `${{ needs.read-version.outputs.c2pa-native-version }}`?

c2pa-native-version.txt
```diff
@@ -0,0 +1 @@
+c2pa-v0.55.0
```
ok-nick, line 1: Another idea is also using `grep` on the `Cargo.toml` or since we have Python scripts setup, we can also parse the toml and get the version defined in our Rust dependencies!

tmathern: ... There is no Cargo.toml in this repo anymore.

.github/workflows/build.yml
tmathern: Indeed, you caught my debug code. Fix incoming.

c2pa-native-version.txt
ok-nick: Didn't know that, looks good!

PR #115, opened 2025-06-06
fix: Test coverage

alextrnnn:
- Add more unit tests (non-threaded)


tests/test_unit_tests.py
```diff
@@ -54,6 +54,17 @@ def test_stream_read_and_parse(self):
             title = manifest_store["manifests"][manifest_store["active_manifest"]]["title"]
             self.assertEqual(title, "C.jpg")
 
+    def test_stream_read_string_stream(self):
+        reader = Reader("image/jpeg", self.testPath)
```
tmathern: As an aside: The tests use `Reader` without releasing the memory resources (without calling close). In tests it's tolerable, but we may want to use something like this in all Reader tests (with statement), to clean up resources:

```
with open(self.testPath, "rb") as file:
    reader = Reader("badFormat", file)
```

Would you mind making that change on the added tests test_stream_read_string_stream and test_stream_read_string_stream_and_parse?

tests/test_unit_tests.py
```diff
@@ -91,6 +102,13 @@ def test_reader_close_cleanup(self):
             self.assertIsNone(reader._own_stream)
             # Verify reader is marked as closed
             self.assertTrue(reader._closed)
+    
+    def test_resource_to_stream_on_closed_reader(self):
```
tmathern, line 106: That test is great as is (don't use the with statement) as we exactly want to test that case for the close call here.

tests/test_unit_tests.py
```diff
@@ -395,6 +429,55 @@ def test_builder_sign_with_ingredient(self):
 
         builder.close()
 
+    def test_builder_sign_with__duplicate_ingredient(self):
```
tmathern:
```suggestion
    def test_builder_sign_with_duplicate_ingredient(self):
```
(Typo)

tmathern (changes requested):
- Use the with statement on Reader in test_stream_read_string_stream and test_stream_read_string_stream_and_parse (also ensure more test diversity, so we cover more ways a developer may want to use the SDK)
- A typo to fix
- Rest looks good, thank you for diving into this!

tmathern (approved):

PR #118, opened 2025-06-17
docs: Update examples (with docs)

tmathern:
- Update examples (with docs)
- Add readme in the examples folder and link it to the examples docs

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -2,18 +2,20 @@
 
 This package works with media files in the [supported formats](https://github.com/contentauth/c2pa-rs/blob/main/docs/supported-formats.md).
 
+> **Note**: For complete working examples, see the [examples folder](https://github.com/contentauth/c2pa-python/tree/main/examples) in the repository.
+
 ## Import
 
-Import the API as follows:
+Import the API:
 
 ```py
-from c2pa import *
+from c2pa import Builder, Reader, Signer, C2paSigningAlg, C2paSignerInfo
 ```
 
 ## Define manifest JSON
 
 The Python library works with both file-based and stream-based operations.
-In both cases, the manifest JSON string defines the C2PA manifest to add to an asset; for example:
+In both cases, the manifest JSON string defines the C2PA manifest to add to an asset. For example:
 
 ```py
 manifest_json = json.dumps({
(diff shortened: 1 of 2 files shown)


docs/usage.md
```diff
@@ -34,58 +36,33 @@ manifest_json = json.dumps({
...
-
-  # Get the active manifest.
-  manifest = reader.get_active_manifest()
-  if manifest != None:
-
-    # get the uri to the manifest's thumbnail and write it to a file
-    uri = manifest["thumbnail"]["identifier"]
-    reader.resource_to_file(uri, "thumbnail_v2.jpg") 
+    # Create a reader from a file path
+    with Reader("path/to/media_file.jpg") as reader:
```
tmathern, line 57: @crandmck I don't know if we need to emphasize that we recommend to use the context manager for proper memory clean up?

docs/usage.md
```diff
@@ -98,76 +75,68 @@ except Exception as err:
...
+except Exception as e:
+    print("Failed to sign manifest store: " + str(e))
 ```
 
 ## Stream-based operation
 
 Instead of working with files, you can read, validate, and add a signed manifest to streamed data.  This example code does the same thing as the file-based example.
 
-### Read and validate C2PA data
+### Read and validate C2PA data using streams
```
tmathern, line 123: To avoid duplicate titles in the md file. There probably are better wordings.

crandmck: Duplicate titles are not a problem.  But this change is fine--just wordier.

docs/usage.md
crandmck: Yes, please mention it in the text, rather than in a code comment.

docs/usage.md
```diff
@@ -2,18 +2,20 @@
...
 ```py
-from c2pa import *
+from c2pa import Builder, Reader, Signer, C2paSigningAlg, C2paSignerInfo
 ```
 
 ## Define manifest JSON
 
 The Python library works with both file-based and stream-based operations.
-In both cases, the manifest JSON string defines the C2PA manifest to add to an asset; for example:
+In both cases, the manifest JSON string defines the C2PA manifest to add to an asset. For example:
```
crandmck, line 21: Do you have something against semicolons?

tmathern: Not specially? I just find it nicer with dots?

docs/usage.md
tmathern: Done, on top (after the import example).

docs/usage.md
```diff
@@ -2,18 +2,22 @@
 
 This package works with media files in the [supported formats](https://github.com/contentauth/c2pa-rs/blob/main/docs/supported-formats.md).
 
+> **Note**: For complete working examples, see the [examples folder](https://github.com/contentauth/c2pa-python/tree/main/examples) in the repository.
```
crandmck: Sorry I should have noted this before.  This kind of formatting doesn't work well on the doc site (which incorporates these files.).  Just remove `> **Note**:`. I think the sentence speaks for itself without a callout.

docs/usage.md
```diff
@@ -2,18 +2,22 @@
...
 
-Import the API as follows:
+Import the objects needed from the API:
 
 ```py
-from c2pa import *
+from c2pa import Builder, Reader, Signer, C2paSigningAlg, C2paSignerInfo
 ```
 
+> **Note**: Both `Builder` and `Reader` support being used as context managers (with `with` statements). It's recommended to use them with context managers to ensure proper resources and memory cleanup.
```
crandmck: Same comment as above re formatting.

Suggest rewording as follows:

You can use both `Builder` and `Reader` classes as context managers by using a `with` statement. Doing this is recommended to ensure proper resource and memory cleanup.

crandmck (commented):
Just a couple more small editorial comments.

docs/usage.md
tmathern: Duplicate titles are flagged by my `.md` linter ;) .

docs/usage.md
tmathern: Done, will be in next swt of pushed changes.

PR #119, opened 2025-06-17
docs: Move project contributors info into the correct file.

crandmck:
Other misc cleanup.

### DIFF
--- a/docs/project-contributions.md
+++ b/docs/project-contributions.md
@@ -7,7 +7,6 @@ The information in this page is primarily for those who wish to contribute to th
 It is best to [set up a virtual environment](https://virtualenv.pypa.io/en/latest/installation.html) for development and testing:
 
 ```bash
-# Create virtual environment
 python -m venv .venv
 ```
 
@@ -46,32 +45,7 @@ This command:
 - Copies the appropriate libraries for your platform from `artifacts/` to `src/c2pa/libs/`
 - Installs the package in development mode, so you can make changes to the Python code without reinstalling.
 
-## Build from source
-
-To build from source on Linux, install `curl` and `rustup` then set up Python.
-
-First update `apt` then (if needed) install `curl`:
-
-```bash
-apt update
-apt install curl
-```
-
-Install Rust:
-
-```bash
-curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
-source "$HOME/.cargo/env"
-```
-
-Install Python, `pip`, and `venv`:
-
-```bash
-apt install python3
-apt install pip
-apt install python3.11-venv
-python3 -m venv .venv
-```
+## Building wheels
 
 Build the wheel for your platform (from the root of the repository):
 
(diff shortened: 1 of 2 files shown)


docs/project-contributions.md
```diff
@@ -2,9 +2,51 @@
...
+```bash
+pip install -e .
+```
+
+This command:
+
+- Copies the appropriate libraries for your platform from `artifacts/` to `src/c2pa/libs/`
+- Installs the package in development mode, so you can make changes to the Python code without reinstalling.
+
+## Build from source
```
crandmck: Is this section still needed?

docs/project-contributions.md
```diff
@@ -42,49 +84,61 @@ pip install -U pytest
 python3 -m build --wheel
 ```
 
-Note: To peek at the Python code (uniffi generated and non-generated), run `maturin develop` and look in the c2pa folder.
+## Building wheels
```
crandmck: The above lines (76-85) seems to duplicate this.  Which is correct?

docs/project-contributions.md
```diff
@@ -42,49 +84,61 @@ pip install -U pytest
...
+├── pyproject.toml            # Python project configuration
+├── requirements.txt          # Python dependencies
+├── requirements-dev.txt      # Development dependencies
+└── setup.py                  # Package setup script
 ```
 
 ## Testing
 
-We use [PyTest](https://docs.pytest.org/) and [unittest](https://docs.python.org/3/library/unittest.html) for testing.
+The project uses [PyTest](https://docs.pytest.org/) and [unittest](https://docs.python.org/3/library/unittest.html) for testing.
```
crandmck, line 97: Do we still use unittest?

docs/project-contributions.md
```diff
@@ -42,49 +84,61 @@ pip install -U pytest
...
+    ```bash
+    make test
+    ```
+5. Alternatively, install pytest (if not already installed) and run it:
+    ```bash
+    pip install pytest
+    pytest
+    ```
 
 ### Testing during bindings development
```
crandmck, line 115: Is this still accurate?

docs/project-contributions.md
tmathern: No, it isn't anymore, good catch.

If you feel something should be mentioned for building everything yourself, it has been replaced by running https://github.com/contentauth/c2pa-python/blob/main/Makefile#L20 followed by https://github.com/contentauth/c2pa-python/blob/main/Makefile#L100 and then https://github.com/contentauth/c2pa-python/blob/main/Makefile#L26. But the shortcut command is mentioned in another part of the docs already, the one you're moving around.

docs/project-contributions.md
tmathern: None. Running https://github.com/contentauth/c2pa-python/blob/main/Makefile#L49 does that locally. https://github.com/contentauth/c2pa-python/blob/main/Makefile#L85 verifies the builds.

But the release builds are only done by CI/CD now.

tmathern: (So the whole section likely can go - we could mention that you can test the wheel builds correctly locally by running the two commands mentioned above)

docs/project-contributions.md
tmathern: That is a great question.

Yes, I use both, and will continue to do so.
pytest has some troubles with virtual envs contexts. It is often useful to check if both fail to find out if it's a virtualenv issue or a "real" code issue. (Basically I am using unittest to sometimes debug pytest - that's why both are here).

docs/project-contributions.md
tmathern: What you left/have updated is correct. We can run test both with pytest or unittest. (`make test` uses unittest).
I recommend using `make test`, to avoid weird environment switch issues that pytest can have if you switch often between multiple virtual environments.

tmathern (approved):

Issue #117, opened 2025-06-17
Could not find libc2pa_c.so in any of the search paths on arm64 architecture


mullevik:
Hello, after you have closed #101, I tried to add `c2pa-python==0.10.0` to my CI workflow.

I am getting an import error specifically on an AWS graviton instance (ARM64 based).
```
RuntimeError: Could not find libc2pa_c.so in any of the search paths
```

Here is a minimal example representing my CI Dockerfile:
```
FROM python:3.12-slim-bookworm

RUN apt-get update
RUN pip install --upgrade pip

RUN pip install c2pa-python==0.10.0

CMD python -c "import c2pa; print('c2pa imported')"
```
- this dockerfile works fine on `x86` architecture
- this dockerfile reproduces the error when running on `arm64` architecture

I have limited knowledge about rust and cpu architectures so I am not sure how to deal with this problem or what actually causes it. Is there anything I can do to fix it on my end?

> Side note: 
> 
> I also have a macos machine with arm64 architecture where the `c2pa-python==0.10.0` package just works natively. So I guess there is not a big issue with the architecture itself, just with the build process for linux.

tmathern:
It doesn't work because we do not support Linux ARM yet. We need to add that to the native build matrix for it to work.

mullevik:
Thanks for the quick reply. I guess I will wait, then. Thanks for the info.

tmathern:
The change was merged and released - https://pypi.org/project/c2pa-python/0.11.0/.

PR #121, opened 2025-06-17
feat: Add a runner for Linux ARM support

tmathern:
https://github.com/contentauth/c2pa-python/issues/117

- Add build steps for Linux ARM (64).


.github/workflows/build-wheel.yml
```diff
@@ -63,7 +63,7 @@ jobs:
             cd /io &&
             /opt/python/cp310-cp310/bin/pip install -r requirements.txt -r requirements-dev.txt &&
             /opt/python/cp310-cp310/bin/pip install toml &&
-            C2PA_LIBS_PLATFORM=\"${{ inputs.architecture == 'aarch64' && 'aarch64-unknown-linux-gnu' || 'x86_64-unknown-linux-gnu' }}\" /opt/python/cp310-cp310/bin/python scripts/download_artifacts.py $C2PA_VERSION &&
+            C2PA_LIBS_PLATFORM=\"${{ format('{0}', inputs.architecture == 'aarch64' && 'aarch64-unknown-linux-gnu' || 'x86_64-unknown-linux-gnu') }}\" /opt/python/cp310-cp310/bin/python scripts/download_artifacts.py $C2PA_VERSION &&
```
tmathern, line 66: Cleverness to inject the target arch we need.

.github/workflows/build.yml
```diff
@@ -51,7 +51,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ macos-latest, ubuntu-latest ]
+        os: [ macos-latest, ubuntu-latest, ubuntu-24.04-arm ]
```
tmathern, line 54: Technically `ubuntu-24.04-arm` is still in public preview, but it seems to work.

gpeacock: We used to do a [manylinux](https://github.com/pypa/manylinux) build that magically supports all kinds of linux version and distributions. It would be nice to get back to that somehow.

gpeacock (approved):

.github/workflows/build.yml
tmathern: Totally agreed.

Issue #122, opened 2025-06-18
Calling Builder.sign_file results in undefined symbol: c2pa_builder_sign_file


rokclimb15:
```
c2pa_signer_create
[ERROR] AttributeError: /usr/local/lib/python3.12/site-packages/c2pa/libs/libc2pa_c.so: undefined symbol: c2pa_builder_sign_file
Traceback (most recent call last):
  File "/index.py", line 96, in lambda_handler
    signed_manifest_bytes = c2pa_lib.write(file_path, manifest_payload, public_key, sign)
  File "/c2pa_lib.py", line 17, in write
    signed_bytes = builder.sign_file(filename, tmp_output, signer)
  File "/usr/local/lib/python3.12/site-packages/c2pa/c2pa.py", line 1812, in sign_file
    result = _lib.c2pa_builder_sign_file(
  File "/usr/local/lib/python3.12/ctypes/__init__.py", line 392, in __getattr__
    func = self.__getitem__(name)
  File "/usr/local/lib/python3.12/ctypes/__init__.py", line 397, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
```

tmathern:
Can you share additional information to help investigate?

Like: Operating system (OS), OS Version, CPU architecture.
Also, are you running in DOcker or in any cloud architecture?

rokclimb15:
This is Debian buster, ARM64. It's running on AWS Lambda via a Docker
image. I suspect this will reproduce on any Linux architecture however, it
looks like a missing exported symbol from the Rust C bindings which are
embedded in this version.

tmathern:
As an aside, I don't think it reproduces on all architectures, because the tests on Ubuntu and Ubuntu ARM for the publish workflows pass, eg https://github.com/contentauth/c2pa-python/pull/121. 

There was a v0.10.0 that had a similar issue. Are you on v0.11.0? That one added more ARM support.

rokclimb15:
Yes, I am using v0.11.0

rokclimb15:
The unit tests are passing on Ubuntu ARM64, but they don't seem to cover
Builder.sign_file. They only cover the deprecated c2pa.sign_file function.
https://github.com/contentauth/c2pa-python/blob/main/tests/test_unit_tests.py#L911

I think that untested code path is where the issue is

rokclimb15:
Thanks, no rush on our end, we're good on 0.6.0 for now.

tmathern:
[...]

PR #123, opened 2025-06-19
fix: Missing symbol from native lib

tmathern:
https://github.com/contentauth/c2pa-python/issues/122

Fix missing symbol (+add missing test coverage).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "c2pa-python"
-version = "0.11.0"
+version = "0.11.1"
 requires-python = ">=3.10"
 description = "Python bindings for the C2PA Content Authenticity Initiative (CAI) library"
 readme = { file = "README.md", content-type = "text/markdown" }


src/c2pa/c2pa.py
```diff
@@ -1725,19 +1725,20 @@ def to_archive(self, stream: Any):
                 raise C2paError(
                     self._error_messages['archive_error'].format("Unknown error"))
 
-    def sign(
+    def _sign_internal(
```
tmathern, line 1728: Refactoring

dyro (approved):

src/c2pa/c2pa.py
```diff
@@ -1795,41 +1821,27 @@ def sign_file(self,
         Args:
             source_path: Path to the source file
             dest_path: Path to write the signed file to
+            signer: The signer to use
 
         Returns:
             A tuple of (size of C2PA data, optional manifest bytes)
```
gpeacock: why are we returning the tuple here? is there some reason we can't just get the size from the bytes?

src/c2pa/c2pa.py
```diff
@@ -1777,12 +1775,40 @@ def sign(
                 manifest_bytes = bytes(manifest_bytes_ptr[:size])
                 _lib.c2pa_manifest_bytes_free(manifest_bytes_ptr)
 
-            return manifest_bytes
+            return result, manifest_bytes
```
gpeacock: I don't know why we need to return the size here. I think we should remove that and always just return the manifest_bytes. The caller can always get the length of the bytes, right?

src/c2pa/c2pa.py
tmathern: I returned the tuple to verify it in some tests. I can remove that...

src/c2pa/c2pa.py
tmathern: Technically no, not if you set to "no_embed" on the builder. It would need a "real" roundtrip to get the written size (which I simulate in tests). But we have other tests covering that, so simplifying is fine.

tests/test_unit_tests_threaded.py
```diff
@@ -1068,109 +1068,6 @@ def read_manifest(reader_id):
         # Verify all readers completed
         self.assertEqual(active_readers, 0, "Not all readers completed")
 
-    def test_remote_sign_threaded(self):
```
tmathern, line 1071: Removed to simplify, since covered by other tests (and the test logic changes due to the signature change, so leaving that to the other tests to check).

src/c2pa/c2pa.py
```diff
@@ -1770,66 +1768,72 @@ def sign(
...
-            manifest_bytes = None
             if manifest_bytes_ptr:
-                # Convert the manifest bytes to a Python bytes object
-                size = result
-                manifest_bytes = bytes(manifest_bytes_ptr[:size])
+                # Free the manifest bytes pointer if it was allocated
                 _lib.c2pa_manifest_bytes_free(manifest_bytes_ptr)
 
-            return manifest_bytes
+            return result
```
gpeacock, line 1775: how are we going to return the manifest bytes if we throw them away here?

src/c2pa/c2pa.py
```diff
@@ -1770,66 +1768,72 @@ def sign(
...
         """Sign a file and write the signed data to an output file.
 
         Args:
             source_path: Path to the source file
             dest_path: Path to write the signed file to
+            signer: The signer to use
 
         Returns:
-            A tuple of (size of C2PA data, optional manifest bytes)
+            Size of C2PA data
```
gpeacock, line 1819: Sign_file needs to return manifest_bytes so that they can be used for sidecar or uploading remote manifests.

tests/test_unit_tests.py
```diff
@@ -709,12 +709,46 @@ def test_builder_set_remote_url_no_embed(self):
...
+
+        # Create a temporary directory for the test
+        temp_dir = tempfile.mkdtemp()
+        try:
+            # Create a temporary output file path
+            output_path = os.path.join(temp_dir, "signed_output.jpg")
+
+            # Use the sign_file method
+            builder = Builder(self.manifestDefinition)
+            result = builder.sign_file(
```
gpeacock, line 731: This result should be manifest_bytes. Errors should throw.

tests/test_unit_tests.py
```diff
@@ -322,13 +322,13 @@ def test_remote_sign(self):
             builder = Builder(self.manifestDefinition)
             builder.set_no_embed()
```
gpeacock, line 323: typically remote signing would also set_remote_url() If remote fetch is disabled the reader will return an error indicating the RemoteUrl that should be read from. Setting no_embed on its own is mainly there for writing sidecars.

src/c2pa/c2pa.py
gpeacock: What is the purpose of returning the size of the manifest_bytes here?

gpeacock (commented):
It looks like this was already merged, but I there are things that need to be fixed

Issue #124, opened 2025-06-20
v0.11.1 builder.sign_file results in invalid signature value


rokclimb15:
I have modified our code just slightly for the new 0.10.0+ methods, but it fails our local test suite which worked under 0.6.0

```
def write(filename: str, manifest: dict, public_key: str, signing_function: Callable) -> tuple[int, bytes | None]:
    manifest_json = json.dumps(manifest)

    builder = c2pa.Builder(manifest_json)

    signer = c2pa.Signer.from_callback(
        signing_function,
        c2pa.C2paSigningAlg.ES256,
        public_key, 
        "http://timestamp.digicert.com"
    )

    tmp_output = os.path.join('/tmp', 'tmp.' + os.path.splitext(filename)[1])

    signed_bytes = builder.sign_file(filename, tmp_output, signer)

    os.rename(tmp_output, filename)

    return signed_bytes
```

```
c2pa_signer_create
c_callback: signed_size: 0
[ERROR] C2paError: Signature: invalid signature value
Traceback (most recent call last):
  File "/index.py", line 96, in lambda_handler
    signed_manifest_bytes = c2pa_lib.write(file_path, manifest_payload, public_key, sign)
  File "/c2pa_lib.py", line 17, in write
    signed_bytes = builder.sign_file(filename, tmp_output, signer)
  File "/usr/local/lib/python3.12/site-packages/c2pa/c2pa.py", line 1836, in sign_file
    return self._sign_internal(signer, mime_type, source_stream, dest_stream)
  File "/usr/local/lib/python3.12/site-packages/c2pa/c2pa.py", line 1769, in _sign_internal
    raise C2paError(error)
```

rokclimb15:
I am using KMS as a remote signer, so I don't have access to the private
key that corresponds. I can try a minimal testcase with a local signing
callback that just returns bytes, I imagine it will also fail.

tmathern:
That would be helpful to repro the issue though. And then we could add it to the lib's test cases and use it to debug.

rokclimb15:
I was not able to get a working local script, the process is just too different from KMS to mimic. However, I noticed that `Builder.sign_file` now returns the number of bytes signed rather than the signed bytes themselves, which our workflow relies upon.

Is there a different set of APIs I can call to use a custom signer and get back the signed bytes? If not, we're stuck on 0.6.0 for now.

tmathern:
[...]

PR #125, opened 2025-06-20
fix: Callback signer

tmathern:
For https://github.com/contentauth/c2pa-python/issues/124

- Fix callback signers bindings
- Builder.sign (and sign_file) return manifest bytes and status (it is quite useful for testing too, and for those who use the returned signed bytes)
- Clean up error message to move them up on class level so they're not recreated with each instance
- Add unit test: ⚠️ It seems the cryptography APIs have changed, so callback function is written accordingly ⚠️ 

For future reference: To get debug logs in-between languages: Log everything that can be logged - Python logger may be helpful to actually see the logs, combined with logging to stderr and flushing logs on a regular basis. Example in https://github.com/contentauth/c2pa-python/pull/125/commits/8e791d1f27bbed683d6ddae942dfd18209618d5e.


tests/test_unit_tests.py
```diff
@@ -748,6 +752,71 @@ def test_sign_file(self):
             # Clean up the temporary directory
             shutil.rmtree(temp_dir)
 
+    def test_sign_file_callback_signer(self):
```
tmathern, line 786: Original failure on run:
```
ERROR: test_sign_file_callback_signer (__main__.TestBuilder.test_sign_file_callback_signer)
Test signing a file using the sign_file method.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/mathern/Desktop/git/@CAI/c2pa-python/./tests/test_unit_tests.py", line 800, in test_sign_file_callback_signer
    result = builder.sign_file(
             ^^^^^^^^^^^^^^^^^^
  File "/Users/mathern/Desktop/git/@CAI/c2pa-python/src/c2pa/c2pa.py", line 1836, in sign_file
    return self._sign_internal(signer, mime_type, source_stream, dest_stream)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mathern/Desktop/git/@CAI/c2pa-python/src/c2pa/c2pa.py", line 1769, in _sign_internal
    raise C2paError(error)
c2pa.c2pa.C2paError: Signature: invalid signature value

----------------------------------------------------------------------
```

tests/test_unit_tests.py
```diff
@@ -18,9 +18,12 @@
 from unittest.mock import mock_open, patch
 import ctypes
 import warnings
+from cryptography.hazmat.primitives import hashes, serialization
```
tmathern, line 21: For the test with callback signer

tests/test_unit_tests.py
```diff
@@ -34,7 +37,7 @@
 
 class TestC2paSdk(unittest.TestCase):
     def test_sdk_version(self):
-        self.assertIn("0.55.0", sdk_version())
+        self.assertIn("0.57.0", sdk_version())
```
tmathern: This needs the new build from https://github.com/contentauth/c2pa-rs/pull/1181 to work

src/c2pa/c2pa.py
```diff
@@ -679,6 +679,67 @@ def sign_file(
             signer.close()
 
 
+def sign_file_with_callback_signer(
+    source_path: Union[str, Path],
+    dest_path: Union[str, Path],
+    manifest: str,
+    signer: 'Signer'
```
tmathern: For this one, the Signer should be created outside the function, so the caller totally handles and owns the callback.

src/c2pa/c2pa.py
```diff
@@ -1308,40 +1364,83 @@ def from_callback(
...
+            'invalid_certs': "Invalid certificate data: {}",
+            'invalid_tsa': "Invalid TSA URL: {}",
+            'encoding_error': "Invalid UTF-8 characters in input: {}"
+        }
+
         # Validate inputs before creating
         if not certs:
             raise C2paError(
-                cls._error_messages['invalid_certs'].format("Missing certificate data"))
+                error_messages['invalid_certs'].format("Missing certificate data"))
```
tmathern: Update to avoid using cls which was not always available.

src/c2pa/c2pa.py
```diff
@@ -1308,40 +1364,83 @@ def from_callback(
...
 
         if tsa_url and not tsa_url.startswith(('http://', 'https://')):
             raise C2paError(
-                cls._error_messages['invalid_tsa'].format("Invalid TSA URL format"))
+                error_messages['invalid_tsa'].format("Invalid TSA URL format"))
 
         # Create a wrapper callback that handles errors and memory management
-        def wrapped_callback(data: bytes) -> bytes:
+        def wrapped_callback(context, data_ptr, data_len, signed_bytes_ptr, signed_len):
```
tmathern: Note that this is very dependent on the native code and its error detection.

src/c2pa/c2pa.py
```diff
@@ -593,26 +592,50 @@ def read_file(path: Union[str, Path],
...
+) -> str:
+    """Sign a file with a C2PA manifest using a signer.
+
+    .. deprecated:: 0.10.0
+        This function is deprecated and will be removed in a future version.
+        Please use the Builder class for signing and the Reader class for reading signed data instead.
+    """
+    ...
+
+def sign_file(
```
tmathern, line 636: Refactored with `overload`.

src/c2pa/c2pa.py
```diff
@@ -1895,29 +1990,7 @@ def create_signer(
         C2paError: If there was an error creating the signer
         C2paError.Encoding: If the certificate data or TSA URL contains invalid UTF-8 characters
     """
-    try:
```
tmathern, line 1898: Refactored away to reuse the class methods in Signer!

src/c2pa/c2pa.py
```diff
@@ -1932,20 +2005,7 @@ def create_signer_from_info(signer_info: C2paSignerInfo) -> Signer:
     Raises:
         C2paError: If there was an error creating the signer
     """
-    signer_ptr = _lib.c2pa_signer_from_info(ctypes.byref(signer_info))
```
tmathern, line 1935: Refactored away to reuse the class methods in Signer!

src/c2pa/c2pa.py
```diff
@@ -1995,12 +2055,11 @@ def ed25519_sign(data: bytes, private_key: str) -> bytes:
     'Reader',
     'Builder',
     'Signer',
-    'version',
     'load_settings',
     'read_file',
     'read_ingredient_file',
     'sign_file',
     'format_embeddable',
-    'ed25519_sign',
```
tmathern, line 2004: Not used in tests anymore

src/c2pa/c2pa.py
```diff
@@ -1308,40 +1328,91 @@ def from_callback(
...
+                error_messages['invalid_certs'].format("Missing certificate data"))
 
         if tsa_url and not tsa_url.startswith(('http://', 'https://')):
             raise C2paError(
-                cls._error_messages['invalid_tsa'].format("Invalid TSA URL format"))
+                error_messages['invalid_tsa'].format("Invalid TSA URL format"))
 
         # Create a wrapper callback that handles errors and memory management
-        def wrapped_callback(data: bytes) -> bytes:
+        def wrapped_callback(
```
tmathern, line 1371: This fixes callback issues betwwen native code and Python

src/c2pa/c2pa.py
```diff
@@ -675,7 +699,7 @@ def sign_file(
         # Ensure resources are cleaned up
         if 'builder' in locals():
             builder.close()
-        if 'signer' in locals():
+        if 'signer' in locals() and own_signer:
```
tmathern, line 720: Only clean up Signer if we created it for this function call only (otherwise the SIgner is owned by the calling code)

src/c2pa/c2pa.py
```diff
@@ -641,25 +659,31 @@ def sign_file(
...
+                dest_stream = Stream(dest_file)
 
-            # Read the signed manifest from the destination file
-            with Reader(dest_path) as reader:
-                return reader.json()
+                # Use the builder's internal signing logic to get manifest bytes
+                result, manifest_bytes = builder._sign_internal(
+                    signer, mime_type, source_stream, dest_stream)
+
+                return manifest_bytes
```
tmathern, line 692: In case the caller needs the raw manifest bytes (instead of a manifest store as JSON)

tmathern:
https://github.com/contentauth/c2pa-python/issues/124

tmathern:
https://github.com/contentauth/c2pa-python/issues/124

src/c2pa/c2pa.py
```diff
@@ -1308,40 +1328,91 @@ def from_callback(
...
+                    return 0
+
+                # Call the user's callback
+                signature = callback(data)
+                if not signature:
+                    # Error: empty signature, native code will handle that too!
+                    return 0
+
+                # Copy the signature back to the C buffer (since callback is
+                # sued in native code)
```
gpeacock: used

src/c2pa/c2pa.py
```diff
@@ -1308,40 +1328,91 @@ def from_callback(
...
+                actual_len = min(len(signature), signed_len)
+                for i in range(actual_len):
```
gpeacock: there's a spiffier way to do this, I think in the stream read code

tmathern: Done.

src/c2pa/c2pa.py
```diff
@@ -593,45 +607,64 @@ def read_file(path: Union[str, Path],
...
+    """
+    ...
+
+@overload
+def sign_file(
+    source_path: Union[str, Path],
+    dest_path: Union[str, Path],
+    manifest: str,
+    signer: 'Signer',
+    return_manifest_as_bytes: bool = False
```
tmathern, line 629: To return the json or the manifest bytes, caller's choice. Defaults to the JSON.

src/c2pa/c2pa.py
```diff
@@ -593,45 +607,64 @@ def read_file(path: Union[str, Path],
...
-        "Please use the Builder class for signing and the Reader class for reading signed data instead.",
-        DeprecationWarning,
-        stacklevel=2
-    )
 
     try:
-        # Create a signer from the signer info
-        signer = Signer.from_info(signer_info)
+        # Determine if we have a signer or signer info
+        if isinstance(signer_or_info, C2paSignerInfo):
```
tmathern, line 664: That (the isinstance check) is due to the overload decorator to get overloads.

src/c2pa/c2pa.py
```diff
@@ -1308,40 +1343,95 @@ def from_callback(
             C2paError: If there was an error creating the signer
             C2paError.Encoding: If the certificate data or TSA URL contains invalid UTF-8 characters
         """
+        # Define error messages locally since they're instance attributes
+        error_messages = {
```
tmathern: Minor centralization

src/c2pa/c2pa.py
```diff
@@ -593,45 +607,64 @@ def read_file(path: Union[str, Path],
...
+    """Sign a file with a C2PA manifest using a signer.
+    """
+    ...
+
+def sign_file(
+    source_path: Union[str, Path],
+    dest_path: Union[str, Path],
+    manifest: str,
+    signer_or_info: Union[C2paSignerInfo, 'Signer'],
+    return_manifest_as_bytes: bool = False
```
gpeacock, line 641: I don't understand what is going on with this function. I think we should get rid of it entirely. But if someone still needs it, it should only be supported for backwards compatibility. Why are we adding new functionality here? We should be moving everything to Builder/Reader. The only reason I kept this was because it supported writing to data_dir, but that is removed here.

src/c2pa/c2pa.py
```diff
@@ -701,18 +740,17 @@ def __init__(self, file):
...
         """
-        # Initialize _closed first to prevent AttributeError during garbage collection
+        # Initialize _closed first to prevent AttributeError
+        # during garbage collection
         self._closed = False
         self._initialized = False
         self._stream = None
 
-        # Generate unique stream ID with timestamp
-        timestamp = int(time.time() * 1000)  # milliseconds since epoch
```
tmathern, line 710: time() does a syscall, so changed that to remove the need for that (since streams are on the critical path of execution quite often)

src/c2pa/c2pa.py
```diff
@@ -701,18 +740,17 @@ def __init__(self, file):
...
-
+        # Generate unique stream ID efficiently using object ID and counter
         # Safely increment stream ID with overflow protection
         if Stream._next_stream_id >= Stream._MAX_STREAM_ID:
             Stream._next_stream_id = 0  # Reset to 0 if we hit the maximum
-        self._stream_id = f"{timestamp}-{Stream._next_stream_id}"
+        self._stream_id = f"{id(self)}-{Stream._next_stream_id}"
```
tmathern, line 771: time() does a syscall, so changed that to remove the need for that (since streams are on the critical path of execution quite often), using the object's id instead, with the stream id. There is a loop risk on the stream id, but the object id, stream id tuple should together should reduce the collision risks, especially during debug sessions.

tests/test_unit_tests.py
```diff
@@ -748,6 +788,418 @@ def test_sign_file(self):
...
+
+                self.assertIn("signature_info", active_manifest)
+                signature_info = active_manifest["signature_info"]
+                self.assertEqual(signature_info["alg"], self.callback_signer_alg)
+
+        finally:
+            shutil.rmtree(temp_dir)
+
+    def test_sign_file_callback_signer_managed(self):
+        """Test signing a file using the sign_file method with context managers."""
```
gpeacock, line 836: Can you explain what you mean by context managers here?  Perhaps we need a different term to describe what this does in the open source world.

tests/test_unit_tests.py
```diff
@@ -748,6 +788,418 @@ def test_sign_file(self):
...
+                # Verify the signature_info contains the correct algorithm
+                self.assertIn("signature_info", active_manifest)
+                signature_info = active_manifest["signature_info"]
+                self.assertEqual(signature_info["alg"], self.callback_signer_alg)
+
+        finally:
+            shutil.rmtree(temp_dir)
+
+    def test_sign_file_callback_signer_managed_multiple_uses(self):
+        """Test that a signer can be used multiple times with context managers."""
```
gpeacock, line 883: same question here about context managers

gpeacock (approved):

tests/test_unit_tests.py
tmathern: (It's the `with` statement, and everything that happens that is nested under it. It tests the memory handling, especially that things are not mistakenly released too early).

tests/test_unit_tests.py
tmathern: (It's the `with` statement, and everything that happens that is nested under it. It tests the memory handling, especially that things are not mistakenly released too early).

src/c2pa/c2pa.py
tmathern: I added it to handle the overloads: one overload handles creating a signer using C2paSignerInfo, the other a signer directly. The last parameter determines if it returns the manifest store as json, or the signed bytes. 
`it should only be supported for backwards compatibility` <= agreed

PR #127, opened 2025-06-25
[ON-HOLD] feat: Add some filetype guessing capabilities

tmathern:
- Add some very basic filetype guessing capabilities to handle files that are extensionless *but* the recommendation for usage would be to use files with extensions. So there is no doubt about the filetype.


src/c2pa/c2pa.py
```diff
@@ -150,6 +150,75 @@ class C2paSigningAlg(enum.IntEnum):
             ctypes.c_ubyte), ctypes.c_size_t)
 
 
+def _guess_mime_type_using_magic_number(file_path: Union[str, Path]) -> Optional[tuple[str, str]]:
```
tmathern, line 153: This is a failsafe more than anything else.

src/c2pa/c2pa.py
```diff
@@ -1915,7 +1990,7 @@ def sign(
         """Sign the builder's content and write to a destination stream.
 
         Args:
-            format: The MIME type or extension of the content
+            format: The MIME type or extension of the content (extension preferred)
```
gpeacock: We have been moving toward mime-types instead of extensions for long time now. I don't want to reverse that policy. Extensions are allowed here more as a fallback/compatibility thing. 
So if some particular app wants to use extensions, that's ok, but we should not document it as preferred. 
When we move this logic into c2pa-rs, we can convert mime types to extensions for if needed.

tmathern: Ok, docs updated!

tmathern:
> This is a great example of why I want most of the functionality tests in Rust instead of repeating them in every language. The appropriate level of testing in python should just be to validate that a file without an extension can be processed. The rest of the tests should be handled in Rust so that we don't need to duplicate them, and all the test files everywhere. Large test files slow down builds. I'd prefer most of these test files to only live in one place. We may want to create a separate repo for them if they really need to be shared like this.

So hmmm, how do we proceed? Do you want me to remove the test files (and tests, then)? Or grab them from c2pa-rs?
Noting that this becomes obsolete once that code has been pulled down there.

tmathern:
Let's put a pin in that for now - if the guessing moves down soon to the c2pa-rs Rust code SDK, we can heavily simplify that (by just adding the tests to verify the functionality).

PR #128, opened 2025-06-27
fix: Update docs

tmathern:
- Small docs update
- Add a test to be closer to having 1:1 tests for what is in the docs


tests/test_unit_tests.py
```diff
@@ -95,6 +95,13 @@ def test_reader_double_close(self):
             with self.assertRaises(Error):
                 reader.json()
 
+    def test_reader_streams_with_nested(self):
```
tmathern, line 98: Bonus test to be really close to what is in the doc

crandmck (approved):

PR #129, opened 2025-06-28
fix: Docs update

tmathern:
- Add an example that showcases v2 manifest signing + callback signer
- Keep old example (renamed to signer_info) showing signing using signer information and v1 manifest
- Add unit tests for V2 signing

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -8,7 +8,9 @@ The `examples/training.py` script demonstrates how to add a "Do Not Train" asser
 
 ### Signing and Verifying Assets
 
-The `examples/sign.py` script shows how to sign an asset with a C2PA manifest and verify it.
+The `examples/sign.py` script shows how to sign an asset with a C2PA manifest and verify it using a callback signer. Callback signers let you define signing logic, eg. where to load keys from.
+
+The `examples/sign_info.py` script shows how to sign an asset with a C2PA manifest and verify it using a "default" signer created with the needed signer information.
 
 ## Running the Examples
 
@@ -21,6 +23,11 @@ Then you can run the examples with the following commands:
 python examples/training.py
 
 # Run the signing and verification example
+# In this example, signing is done with a Signer created using SignerInfo
+python examples/sign_info.py
+
+# Run the signing and verification example
+# In this example, signing is done using a callback signer
 python examples/sign.py
 ```
 
--- a/examples/sign.py
+++ b/examples/sign.py
@@ -58,15 +58,17 @@ manifest_definition = {
     "ingredients": [],
     "assertions": [
         {
-            'label': 'stds.schema-org.CreativeWork',
-            'data': {
-                '@context': 'http://schema.org/',
-                '@type': 'CreativeWork',
-                'author': [
-                    {'@type': 'Person', 'name': 'Example User'}
+            "label": "c2pa.actions",
+            "data": {
+                "actions": [
+                    {
+                        "action": "c2pa.created",
+                        "parameters": {
+                            # could hold additional information about this step
+                        }
+                    }
                 ]
-            },
-            'kind': 'Json'
+            }
         }
     ]
 }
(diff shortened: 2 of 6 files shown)


examples/sign_info.py
```diff
@@ -0,0 +1,105 @@
+# Copyright 2025 Adobe. All rights reserved.
```
tmathern, line 1: Technically this is the "old" previous version of the sign.py example. Kept for now for reference.

examples/training.py
```diff
@@ -42,25 +42,7 @@
 def getitem(d, key):
     return reduce(operator.getitem, key, d)
 
-
-# This function signs data with PS256 using a private key
-def sign_ps256(data: bytes, key: bytes) -> bytes:
```
tmathern, line 47: Not used in this example, so cleaned up.

examples/sign.py
```diff
@@ -11,79 +11,110 @@
...
         {
-            'label': 'stds.schema-org.CreativeWork',
-            'data': {
-                '@context': 'http://schema.org/',
-                '@type': 'CreativeWork',
-                'author': [
-                    {'@type': 'Person', 'name': 'Example User'}
+            "label": "c2pa.actions",
+            "data": {
+                "actions": [
```
tmathern, line 82: I believe this is the most minimal valid V2 manifest we can have.

examples/sign.py
```diff
@@ -11,79 +11,110 @@
...
+                    {
+                        "action": "c2pa.created",
+                        "parameters": {
```
tmathern, line 85: Optional, left here for informational purposes (see comment)

examples/sign.py
```diff
@@ -11,79 +11,110 @@
...
 
 import os
 import c2pa
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.backends import default_backend
+
+# Note: Builder, Reader, and Signer support being used as context managers
+# (with 'with' statements), but this example shows manual usage which requires
+# explicitly calling the close() function to clean up resources.
```
tmathern, line 24: I left it without the with as I found it to be more easy tor ead through without the nesting, on first read.

examples/README.md
```diff
@@ -8,6 +8,9 @@ The examples use asset files from the `tests/fixtures` directory, save the resul
 The [`examples/sign.py`](https://github.com/contentauth/c2pa-python/blob/main/examples/sign.py) script shows how to sign an asset with a C2PA manifest and verify the asset.
 
 
+The `examples/sign.py` script shows how to sign an asset with a C2PA manifest and verify it using a callback signer. Callback signers let you define signing logic, eg. where to load keys from.
```
crandmck: instead of "eg", spell out "for example".

examples/README.md
```diff
@@ -77,7 +80,17 @@ Run the "do not train" assertion example:
 python examples/training.py
 ```
 
-Run the signing and verification example:
+### Run the signing and verification example
+
+In this example, signing is done with a Signer created using SignerInfo:
```
crandmck: In this example, `SignerInfo` creates a `Signer` object that signs the manifest.

examples/README.md
```diff
@@ -77,7 +80,17 @@ Run the "do not train" assertion example:
...
+
+```bash
+python examples/sign_info.py
+```
+
+### Run the callback signing and verification example
+
+In this example, signing is done using a callback signer:
```
crandmck: In this example, a callback signer does the signer:

examples/sign_info.py
crandmck: I suggest we add a comment noting this, so people don't get confused.  That is, if this is a deprecated examples let's say that.

tmathern: Ack - done.

examples/README.md
tmathern: `In this example, a callback signer is the signer` is more accurate (is instead of does - because it is only one part that we use, we still provide the whole other plumbing to sign)

Issue #135, opened 2025-08-01
C2paSigningAlg cannot be passed to C2paSignerInfo


djfletcher:
The library provides a `C2paSigningAlg` enum which presumably should be passed as the `alg` parameter to the `C2paSignerInfo` object. However, it raises an exception when I do that:
```python
Traceback (most recent call last):
  File "/Users/daniel/dev/c2pa-python/examples/sign.py", line 47, in <module>
    signer = c2pa.Signer.from_info(signer_info)
  File "/Users/daniel/dev/c2pa-python/venv/lib/python3.13/site-packages/c2pa/c2pa.py", line 1332, in from_info
    raise C2paError(error)
c2pa.c2pa.C2paError: NullParameter: signer_info.alg
```

This is the minimum diff to reproduce the error in `examples/sign.py`:
```diff
diff --git a/examples/sign.py b/examples/sign.py
index 82070b4..7d365e3 100644
--- a/examples/sign.py
+++ b/examples/sign.py
@@ -38,7 +38,7 @@ certs = open(fixtures_dir + "es256_certs.pem", "rb").read()
 key = open(fixtures_dir + "es256_private.key", "rb").read()
 
 signer_info = c2pa.C2paSignerInfo(
-    alg=b"es256",  # Use bytes instead of encoded string
+    alg=c2pa.C2paSigningAlg.ES256,
     sign_cert=certs,
     private_key=key,
     ta_url=b"http://timestamp.digicert.com"  # Use bytes and add timestamp URL
```

tmathern:
You can still use the "old" way of passing the algorithm value, while I look at this specifically:
```
signer_info = C2paSignerInfo(
  alg=b"es256",
  sign_cert=certs,
  private_key=key,
  ta_url=b"http://timestamp.digicert.com"
)
```

(That is taken directly from the tests, so should work :) ).

tmathern:
This has been fixed in https://github.com/contentauth/c2pa-python/releases/tag/v0.12.1.
Release version on PyPI: https://pypi.org/project/c2pa-python/0.12.1/

PR #136, opened 2025-08-04
fix: Support using Alg enum in SignerInfo

tmathern:
Currently, in the `C2paSignerInfo` class, we expected the algorithm string as bytes.

As reported in https://github.com/contentauth/c2pa-python/issues/135, I believe it makes sense to allow to use the enum and that we parse it accordingly into the value needed by the native call (@gpeacock At some point we should harmonize that API - so that we take the same types everywhere, wdyt?).

This PR adds support for that, and also makes sure the timestamp URL gets properly converted into bytes since natively we expect bytes too (and not just a string).


src/c2pa/c2pa.py
```diff
@@ -205,6 +205,60 @@ class C2paSignerInfo(ctypes.Structure):
         ("ta_url", ctypes.c_char_p),
     ]
 
+    # Mapping from C2paSigningAlg enum to string representation
+    _ALG_MAPPING = {
```
djfletcher: Instead of a mapping, how about making these byte-encoded strings the values of [C2paSigningAlg](https://github.com/contentauth/c2pa-python/blob/423e82bfb20d2abd4af8a6449d43afa47af9d298/src/c2pa/c2pa.py#L115) directly? Like this:

```python

class C2paSigningAlg(enum.Enum):
    """Supported signing algorithms."""
    ES256 = b"es256"
    ES384 = b"es384"
    ES512 = b"es512"
    PS256 = b"ps256"
    PS384 = b"ps384"
    PS512 = b"ps512"
    ED25519 = b"ed25519"
```

tmathern: Have you tried your change in a branch (PRs are always welcome!)? You will likely see tests breaks as the current alg enum values in some calls (most calls - all other calls) needs to be mapped back to an integer for the native calls. Hence the specific mapping here for this case (the less often used mapping enum -> string). I'm adding an explanatory comment though.

src/c2pa/c2pa.py
```diff
@@ -205,6 +218,49 @@ class C2paSignerInfo(ctypes.Structure):
         ("ta_url", ctypes.c_char_p),
     ]
 
+    def __init__(self, alg=None, sign_cert=None, private_key=None, ta_url=None):
```
tmathern: This is a lot of syntactic sugar to make enums reusable and less dependent on underlying (native) types. Also lots of type checkings checks when converting.

tmathern:
Failing tests will be fixed once fix from https://github.com/contentauth/c2pa-rs/pull/1273comes in.

dyro (approved):

PR #140, opened 2025-08-20
WIP: Bump to latest native libs version

tmathern:
- Update to support latest version of native libs 
- Some stream handling changes (due to subtle bugs)
- Refactor ingredient handling from stream
- Fix some typings


tmathern:
https://github.com/contentauth/c2pa-rs/issues/1330

tmathern:
https://github.com/contentauth/c2pa-rs/issues/1331

examples/training.py
```diff
@@ -49,24 +48,41 @@ def getitem(d, key):
         "name": "python_test",
         "version": "0.1"
     }],
+    "claim_version": 2,
```
gpeacock: This should no longer be necessary with the new c2pa_rs api since v2 is now default.

examples/training.py
```diff
@@ -49,24 +48,41 @@ def getitem(d, key):
...
-                "cawg.ai_generative_training": {
-                    "use": "notAllowed"
+    "assertions": [
+        {
+            "label": "c2pa.actions",
+            "data": {
+                "actions": [
+                    {
+                        "action": "c2pa.created",
+                        "parameters": {
```
gpeacock: You must supply a digitalSourceType with a c2pa.created asssertion.

setup.py
```diff
@@ -47,24 +44,15 @@ def get_platform_identifier(cpu_arch = None) -> str:
...
-    - x86_64-apple-darwin (for Mac x86_64)
+    - universal-apple-darwin (for macOS)
     - x86_64-pc-windows-msvc (for Windows 64-bit)
     - x86_64-unknown-linux-gnu (for Linux 64-bit)
     - aarch64-unknown-linux-gnu (for Linux ARM64)
     """
     system = platform.system().lower()
 
     if system == "darwin":
-        if cpu_arch is None:
```
gpeacock, line 60: This is probably ok for Python builds but it will create larger wheels for those who only want one architecture.

tests/benchmark.py
```diff
@@ -37,18 +37,18 @@
...
-                    {'@type': 'Person',
-                        'name': 'Gavin Peacock'
-                     }
+        {
+            "label": "c2pa.actions",
+            "data": {
+                "actions": [
+                    {
+                        "action": "c2pa.created",
+                        "parameters": {
```
gpeacock: c2pa.created requires a digitalSourceType field.

.gitignore
```diff
@@ -114,3 +114,5 @@ target/
 *.dll
 *.so
 src/c2pa/libs/
+!tests/fixtures/*.pem
```
gpeacock, line 117: Do you really want to ignore these? I think they are required for the tests

Makefile
```diff
@@ -20,45 +20,45 @@ clean-c2pa-env: clean
...
 	python3 ./examples/sign.py
+	python3 ./examples/sign_info.py
 	python3 ./examples/training.py
 	rm -rf output/
 
 # Runs the examples, then the unit tests
 test:
-	python3 ./tests/test_unit_tests.py
-	python3 ./tests/test_unit_tests_threaded.py
+	PYTHONWARNINGS=ignore::DeprecationWarning python3 ./tests/test_unit_tests.py
```
gpeacock: What is causing the deprecation errors here? can we update them to avoid this?

src/c2pa/c2pa.py
```diff
@@ -1427,18 +1612,25 @@ def wrapped_callback(
...
+                if (
+                    not data_ptr
+                    or data_len <= 0
+                    or not signed_bytes_ptr
+                    or signed_len <= 0
+                ):
                     # Error: invalid input, invalid so return -1,
                     # native code will handle it!
-                    return -1
+                    return -1  # pragma: no cover
```
gpeacock: We probably should have test cases for these, so this will just hide that they are missing. But it may not be practical to create them. I'm not sure what the goal is here.

src/c2pa/c2pa.py
```diff
@@ -2007,15 +2287,31 @@ def sign_file(self,
...
+            # Open source file and create BytesIO for destination
+            with open(source_path, 'rb') as source_file:
+                source_stream = Stream(source_file)
 
-            # Use the internal stream-base signing logic
-            return self._sign_internal(
-                signer, mime_type, source_stream, dest_stream)
+                # In-memory buffer stream that will be flushed to target,
+                # to ensure proper in-memory processing and validatin
+                dest_stream = Stream(io.BytesIO(bytearray()))
```
gpeacock: Why are we doing this in memory, the API accepts a stream. Is something broken or too slow?

gpeacock (changes requested):

setup.py
tmathern: Yes, but we only also publish universal for now, no? I only see the universal one in https://github.com/contentauth/c2pa-rs/releases/tag/c2patool-v0.20.1 ? And that's the library I have to use for mac currently anyway, lacking alternatives.

.gitignore
tmathern: The `!` unignores them (inverts the logic). One of those was ignored because one of the lines above, so I made sure the test ones aren't.

Makefile
tmathern: It's my tests on the legacy API that log deprecation warnings. I can solve that differently though (to be more precise and even verify the warnings don't happen when they shouldn't).

src/c2pa/c2pa.py
tmathern: Something is maybe broken. The stream if not in-memory fails validation with assertions mismatches. Seems there is some trouble in the streams in how they write/do checks when passing around languages.

src/c2pa/c2pa.py
tmathern: I was looking into it, but I do not have a good (not too long, not too overengineered) way of mocking the library currently to trigger these errors.

examples/training.py
tmathern: Sometimes, I like to be explicit :). But I can comment them out and explain it's the default.

examples/training.py
tmathern: ok, fixed! (In tests & examples).

tests/benchmark.py
tmathern: Fixed!

src/c2pa/c2pa.py
tmathern: Something _was_ broken. Turns out target streams when opened need to be opened in w+b in Python (write, read and binary), not just write binary. Fixing that allowed me to do some refactoring too!

examples/sign.py
```diff
@@ -82,10 +84,8 @@ def callback_signer_es256(data: bytes) -> bytes:
                 "actions": [
                     {
                         "action": "c2pa.created",
-                        "parameters": {
-                            # could hold additional information about this step
-                            # eg. model used, etc.
-                        }
+                        "digitalSourceType": "http://cv.iptc.org/newscodes/digitalsourcetype/digitalCreation",
+                        "softwareAgent": "Python C2PA SDK example"
```
gpeacock: If you want to add softwareAgent, it should be in the new format. It is no longer a String, its the same format as ClaimGeneratorInfo. Version and Icon are optional fields. But it may be simpler to just leave it out. It isn't required here.

"softwareAgent": {
    "name": "Python C2PA SDK example",
    "version": "0.1" 
  }

examples/sign_info.py
```diff
@@ -77,9 +81,8 @@
                 "actions": [
                     {
                         "action": "c2pa.created",
-                        "parameters": {
-                            # could hold additional information about this step
-                        }
+                        "digitalSourceType": "http://cv.iptc.org/newscodes/digitalsourcetype/digitalCreation",
+                        "softwareAgent": "Python C2PA SDK example",
```
gpeacock: see note above on softwareAgent (FYI this was deprecated in v1 claims long ago, so we need to move forward.

examples/sign_info.py
```diff
@@ -93,7 +96,9 @@
 # Sign the image
 print("\nSigning the image...")
 with open(fixtures_dir + "C.jpg", "rb") as source:
-    with open(output_dir + "C_signed.jpg", "wb") as dest:
+    # File needs to be opened in write+read mode to be signed
+    # and verified properly.
+    with open(output_dir + "C_signed.jpg", "w+b") as dest:
```
gpeacock, line 100: ah yes, those little differences that quietly make all the difference!

setup.py
gpeacock: I think it is fine, you know best for Python Pypi publishing now!

gpeacock (approved):
Approved on the condition that you either remove the softwareAgent fields or replace them with structures.

examples/sign.py
tmathern: Ok - I will leave it out then.

tmathern:
> Approved on the condition that you either remove the softwareAgent fields or replace them with structures.

Done: I removed them. Let's keep it simple for the examples and tests.

Issue #138, opened 2025-08-15
c2pa-python 0.12.0 version running error for macOS  x86_64


ttbuffey:
when I upgraded c2pa-python from 0.6.0 to 0.12.0, it show the below error. It seems lacking some lib files for macOS  x86_64, it only support  macOSARM

````
2025-08-15` 13:48:27,306 - ERROR - Platform: universal-apple-darwin, Architecture: **x86_64**
Traceback (most recent call last):
    from c2pa import Builder, C2paSigningAlg as SigningAlg, sign_ps256
  File "/lib/python3.11/site-packages/c2pa/__init__.py", line 7, in <module>
    from .c2pa import (
  File "/lib/python3.11/site-packages/c2pa/c2pa.py", line 103, in <module>
    _lib = dynamically_load_library(_lib_name_default)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.11/site-packages/c2pa/lib.py", line 254, in dynamically_load_library
    raise RuntimeError(f"Could not find {lib_name} in any of the search paths (Platform: {platform_id}, Architecture: {current_arch})")
RuntimeError: Could not find libc2pa_c.dylib in any of the search paths (Platform: universal-apple-darwin, Architecture: x86_64)
````

when I use verison 0.6.0, it worked before.

tmathern:
I will look into updating code and builds to provide an universal mac wheel, which should fix the issue (but it will take a minute). I'll post a PR/WIP PR here too once I start having something.

ttbuffey:
Hi @tmathern , I tried install the specific branch:
`pip install git+https://github.com/contentauth/c2pa-python.git@mathern/switch-to-universal-mac-build`

the same issue occurs, I still didn't see the libc2pa_c.dylib file in the installed site-packages. please check the below screenshot. Is my process correct? the python version is Python 3.11.1

<img width="294" height="475" alt="Image" src="https://github.com/user-attachments/assets/28e9ab06-a283-4638-94ac-d913acd5b560" />

tmathern:
I've been working on https://github.com/contentauth/c2pa-python/pull/148. I ran a test publish to PyPI test (from that PR), a v0.14.0-alpha: https://test.pypi.org/project/c2pa-python/0.14.0a0/ that should have all wheels for mac (arm, Intel, universal). Do you want to give it a try @ttbuffey ?

PR #142, opened 2025-08-22
feat: Python wheel for mac is universal build

tmathern:
- Python wheel for mac is universal build


tmathern:
https://github.com/contentauth/c2pa-python/issues/138

tmathern:
https://github.com/contentauth/c2pa-python/issues/138

tmathern:
The `macos-13` runner is an Intel runner. So I added it in the test matrix to verify the built wheel for macOS also runs well on Intel.

.github/workflows/build-wheel.yml
```diff
@@ -110,8 +118,11 @@ jobs:
           # Build wheel
           python setup.py bdist_wheel --plat-name win_amd64
 
-      - name: Build macOS wheel (Apple Silicon)
-        if: runner.os == 'macOS' && runner.arch == 'arm64'
+          # Verify wheel structure
+          twine check dist/*
+
+      - name: Build macOS wheel (Universal)
```
ok-nick: Just wondering the implications of having separate Intel/Silicon builds? My thought process is that having them separate would presumably reduce build size substantially?

tmathern: Not substantially, around 1MB it seems. Previous release didn't have all the arch builds, now with v0.60.1 I see them. So I'm going to reconsider that change to target the arch (the Intel build was the one missing).

PR #143, opened 2025-08-22
Change return type of signing method to bytes

rokclimb15:
This fixes an issue with static analysis since the function actually returns bytes now


tmathern:
(I'll approve and merge the PR once the CLA is signed and all tests pass).

PR #148, opened 2025-08-27
fix: Wheels for macOS arch (arm, intel, universal)

tmathern:
⚠️ I want to go stepwise here, https://github.com/contentauth/c2pa-python/pull/146 should be merged first, and this one second (in case of reverts) ⚠️ 

- Build macOS wheel for arm, intel and universal (fixes https://github.com/contentauth/c2pa-python/issues/138 better than https://github.com/contentauth/c2pa-python/pull/146 does by providing builds matching CPU architectures, but also introduces more complexity in libs loading and builds).

Test workflow run: https://github.com/contentauth/c2pa-python/actions/runs/17278422499
Test publish to PyPI test seems to work as it should:
Screenshot 2025-08-27 at 14 01 46


.github/workflows/build-wheel.yml
```diff
@@ -135,16 +135,40 @@ jobs:
           pip install wheel
           pip install twine
 
-          # Download native artifacts
-          python scripts/download_artifacts.py $C2PA_VERSION
+          # Download native artifacts for the target architecture
+          if [ "${{ inputs.architecture }}" = "universal2" ]; then
```
gpeacock, line 139: Why is it universal2? Is that a new standard or just some way to avoid conflicts?

gpeacock (approved):
Approved but with question about universal2

.github/workflows/build-wheel.yml
tmathern: It is universal2 because "universal" was when mac did the switch from PowerPC to Intel. So universal flags PowerPC/Intel support. universal2 means Intel/ARM.

PR #151, opened 2025-08-28
fix: DNG handling

tmathern:
mimetype module guesses the wrong mimetype for DNG based on extension, so for `.dng` extension we now bypass the check and set it ourselves.


tests/test_unit_tests.py
```diff
@@ -126,9 +126,6 @@ def test_reader_streams_with_nested(self):
     def test_reader_close_cleanup(self):
         with open(self.testPath, "rb") as file:
             reader = Reader("image/jpeg", file)
-            # Store references to internal objects
```
tmathern, line 129: This one is a minor clean-up of unused variables

src/c2pa/c2pa.py
```diff
@@ -1231,8 +1235,22 @@ def __init__(self,
...
+            # so file_extension will be ""
+            path_obj = Path(path)
+            file_extension = path_obj.suffix.lower() if path_obj.suffix else ""
+
+            if file_extension == ".dng":
+                # mimetypes guesses the wrong type for dng,
+                # so we bypass it and set the correct type
+                mime_type = "image/dng"
+            else:
+                mime_type = mimetypes.guess_type(
```
gpeacock: we should move some common logic for this into the sdk.

tmathern: done: https://github.com/contentauth/c2pa-rs/issues/1357

PR #164, opened 2025-09-16
docs: Add sphinx API docs

crandmck:
Ths PR adds a GH worfkow based on https://github.com/creator-assertions/creator-assertions.github.io/blob/main/.github/workflows/publish.yml, which will build the docs using Sphinx and deploy them to GH Pages for consumption.  We can then link to that from the OS doc site.

Temp online preview of generated docs: https://crandmck.github.io/c2pa-python-docs/

Note:
- The script `scripts/generate_api_docs.py` generates docs into the `api-docs/_build/html` directory (to separate them from the handwritten markdown docs). 
- The Sphinx configuration is in `api-docs/conf.py`
- Info on the API docs is in `docs/project-contributions.md` since (once we set up the automatic doc generation workflow) most users won't need to know how to generate the docs, unless contributing to the project.


src/c2pa/c2pa.py
```diff
@@ -1242,7 +1226,14 @@ def initialized(self) -> bool:
 
 
 class Reader:
-    """High-level wrapper for C2PA Reader operations."""
+    """High-level wrapper for C2PA Reader operations.
+    
+    Example:
+        ```
+        with Reader(path) as reader:
```
tmathern: Technically, we would prefer if people use the stream API instead.
The stream API is like this: `with Reader("image/jpeg", output) as reader:`, where output is either an in-memory stream, or an already oepned file.

src/c2pa/c2pa.py
```diff
@@ -2309,6 +2300,12 @@ def add_ingredient(self, ingredient_json: str, format: str, source: Any):
             C2paError: If there was an error adding the ingredient
             C2paError.Encoding: If the ingredient JSON contains
               invalid UTF-8 characters
+        
+        Example:
+            ```
+            with open(ingredient_file_path, 'rb') as f:
```
tmathern:
```suggestion
            with open(ingredient_file_path, 'rb') as a_file:
```

src/c2pa/c2pa.py
```diff
@@ -2309,6 +2300,12 @@ def add_ingredient(self, ingredient_json: str, format: str, source: Any):
...
+                builder.add_ingredient(ingredient_json, "image/jpeg", f)
```
tmathern:
```suggestion
                builder.add_ingredient(ingredient_json, "image/jpeg", a_file)
```

src/c2pa/c2pa.py
crandmck: Changed the example code accordingly.

src/c2pa/c2pa.py
crandmck: Fixed

src/c2pa/c2pa.py
crandmck: Fixed

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Setup Pages
+        uses: actions/configure-pages@v5
+
+      - uses: ./.github/actions/install-dependencies
+        name: Install dependencies
+
+      - run: pnpm exec nx run-many -t build --projects=tag:lib
+
+      - name: Build docs
+        run: pnpm ci:docs
```
tmathern: Needs to call your Python script here

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,49 @@
...
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          filter: tree:0
+          fetch-depth: 0
+
+      - name: Setup Pages
```
tmathern, line 30: Add the step to setup Python. Exactly like this one: https://github.com/contentauth/c2pa-python/blob/main/.github/workflows/build.yml#L60-L64

tmathern (changes requested):

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,49 @@
+name: Build and publish docs
+
+on:
+  push:
```
tmathern: If you change the trigger, you can test directly on the PR:
```
pull_request:
    types:
      - opened
      - reopened
      - synchronize
      - labeled
 ```

Workflow name will be `Build and publish docs`

tmathern:
@crandmck I configured the repo so you can test publish from this branch. Once you merge, let me know and I'll remove the config. It is also configured to run this workflow from the main branch when a GIthub release is created. I had test deployments that worked, see https://contentauth.github.io/c2pa-python/. Feel free to merge when you are happy with how the docs look. (Note that the branch will need to be up-to-date with latest main to allow merging).

tmathern (approved):

PR #171, opened 2025-09-22
fix: Improve Reader internal code by using with statements

tmathern:
-  Improve Reader internal code by using with statements in more places (perf overhead is minimal, confirmed by banchmark)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,7 @@ publish: release
 
 # Code analysis
 check-format:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -94,6 +94,7 @@ publish: release
 
 # Code analysis
 check-format:
+	python3 -m py_compile src/c2pa/c2pa.py
```
tmathern, line 97: Also add syntax checks.

PR #178, opened 2025-10-03
feat: Add bindings

tmathern:
Add 2 more bindings/helper APIs:
- Add binding for c2pa_reader_is_embedded
- Add binding for c2pa_reader_remote_url


gpeacock (approved):
Thanks, We should generally try to keep all the SDKs aligned like this unless the functionality really doesn't make much sense there.

PR #179, opened 2025-10-03
feat: Add convenience APIs

tmathern:
Code changes:
Some of our other SDKs have convenience methods that help dealing with manifests.
I thought this is a good idea to have too for developer experience, so I added the following, all on the Reader:
- `get_active_manifest`, returns the active manifest (parsed as a dictionary for easy use);
- `get_manifest_by_label`, retrieves any manifest based on its label/id and returns it as dictionary;
- `get_validation_state`, returns just the validation state of the whole manifest store;
- `get_validation_results` returns all the validation results from the manifest store (and only the validation results).

Workflow changes:
- Add formatting and syntax check in github workflow

### DIFF
--- a/src/c2pa/c2pa.py
+++ b/src/c2pa/c2pa.py
@@ -1247,6 +1247,8 @@ class Reader:
         with Reader("image/jpeg", output) as reader:
             manifest_json = reader.json()
             active_manifest = reader.get_active_manifest()
+            validation_state = reader.get_validation_state()
+            validation_results = reader.validation_results()
         ```
         Where `output` is either an in-memory stream or an opened file.
     """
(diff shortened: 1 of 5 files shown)


src/c2pa/c2pa.py
```diff
@@ -1644,6 +1644,122 @@ def json(self) -> str:
...
+        if "manifests" not in manifest_data:
+            raise KeyError("No 'manifests' key found in manifest data")
+
+        manifests = manifest_data["manifests"]
+        if active_manifest_id not in manifests:
+            raise KeyError("Active manifest id not found in manifest store")
+
+        return manifests[active_manifest_id]
+
+    def get_manifest_by_label(self, label: str) -> dict:
```
gpeacock: In Rust we just call this get_manifest() so maybe we should stick with that, or add _from_label. I'd like to keep things as consistent as possible, while allowing for language differences ( like adding a get_  prefix)

src/c2pa/c2pa.py
```diff
@@ -1644,6 +1644,122 @@ def json(self) -> str:
...
+            assertions, ingredients, and signature information.
+
+        Raises:
+            C2paError: If there was an error getting the manifest JSON
+            KeyError: If the active_manifest key is missing from the JSON
+        """
+        # Self-read to get the JSON
+        manifest_json_str = self.json()
+        try:
+            manifest_data = json.loads(manifest_json_str)
```
gpeacock: it may not be worth it since we generally only call these functions once, but we could consider caching the parsed manifest data to avoid generating and parsing it multiple times.

src/c2pa/c2pa.py
```diff
@@ -1644,6 +1644,122 @@ def json(self) -> str:
...
+        except json.JSONDecodeError as e:
+            raise C2paError(f"Failed to parse manifest JSON: {str(e)}") from e
+
+        # Get the active manfiest id/label
```
gpeacock: We got some feedback saying users didn't like the existing exceptions we have, so I'm wondering users would prefer returning None when it fails vs handling the exceptions?

src/c2pa/c2pa.py
tmathern: `get_manifest_from_label` is sold!

src/c2pa/c2pa.py
tmathern: Sold too.

src/c2pa/c2pa.py
tmathern: I updated to return None for the new methods, but did not cross the bridge to change the exception handling for the whole Reader when there is no JUMBF data yet.

tests/test_unit_tests.py
```diff
@@ -270,6 +336,115 @@ def test_read_all_files_using_extension(self):
             except Exception as e:
                 self.fail(f"Failed to read metadata from {filename}: {str(e)}")
 
+    def test_read_cached_all_files(self):
+        """Test reading C2PA metadata with cache functionality from all files in the fixtures/files-for-reading-tests directory"""
+        reading_dir = os.path.join(self.data_dir, "files-for-reading-tests")
+
+        # Map of file extensions to MIME types
+        mime_types = {
```
gpeacock, line 344: I think we should add some exension/mime utilities to the c_ffi_api.

src/c2pa/c2pa.py
```diff
@@ -1642,7 +1680,128 @@ def json(self) -> str:
...
+
+            manifests = manifest_data["manifests"]
+            if active_manifest_id not in manifests:
+                raise KeyError("Active manifest not found in manifest store")
+
+            return manifests[active_manifest_id]
+        except C2paError.ManifestNotFound:
+            return None
+
+    def get_manifest_from_label(self, label: str) -> Optional[dict]:
```
gpeacock: I'd rather have this just be get manifest, since that is what it is called in Rust, but I agree it is very specific this way.

tests/test_unit_tests.py
tmathern: Oh we have them! It's jsut that in the tests I filter further (I'm being clever for as long as I can be by removing for instance files where there is nothing to read as they currently are from a given type). The checks are already done too in Reader/Builder for supported mimetypes, and the functions exposed!

src/c2pa/c2pa.py
tmathern: Sounds good, I can make the change in the name!

PR #181, opened 2025-10-09
feat: add_action API

tmathern:
- Expose the add_action API in Python SDK
- Make sure APIs can also take dict (as convenience), additionally to strings for `add_ingredient_*`, `add_action` and `load_settings`.

### DIFF
--- a/src/c2pa/c2pa.py
+++ b/src/c2pa/c2pa.py
@@ -692,7 +692,7 @@ def load_settings(settings: Union[str, dict], format: str = "json") -> None:
     Raises:
         C2paError: If there was an error loading the settings
     """
-    # If settings is a dict, convert it to JSON string and set format
+    # Convert to JSON string as necessary
     try:
         if isinstance(settings, dict):
             settings_str = json.dumps(settings)
@@ -2573,7 +2573,6 @@ class Builder:
         """
         self._ensure_valid_state()
 
-        # Convert dict to JSON string as necessary
         if isinstance(ingredient_json, dict):
             ingredient_json = json.dumps(ingredient_json)
 
@@ -2665,7 +2664,6 @@ class Builder:
         """
         self._ensure_valid_state()
 
-        # Convert dict to JSON string as necessary
         if isinstance(action_json, dict):
             action_json = json.dumps(action_json)
 


src/c2pa/c2pa.py
```diff
@@ -2613,6 +2618,38 @@ def add_ingredient_from_file_path(
...
+        Raises:
+            C2paError: If there was an error adding the action
+            C2paError.Encoding: If the action JSON contains invalid UTF-8 characters
+        """
+        self._ensure_valid_state()
+
+        try:
+            action_str = action_json.encode('utf-8')
+        except UnicodeError as e:
+            raise C2paError.Encoding(
```
tmathern, line 2673: I think in this case it is justified to throw rather than silently fail. To make sure it is visible the action may not be added.

gpeacock: yes absolutely!

tests/test_unit_tests.py
```diff
@@ -2323,6 +2324,334 @@ def test_builder_state_with_invalid_native_pointer(self):
...
             builder.set_no_embed()
 
+    def test_builder_add_action_to_manifest_no_auto_add(self):
+        # For testing, remove auto-added actions
+        load_settings('{"builder":{"actions":{"auto_placed_action":{"enabled":false}}}}')
+        load_settings('{"builder":{"actions":{"auto_opened_action":{"enabled":false}}}}')
+        load_settings('{"builder":{"actions":{"auto_created_action":{"enabled":false}}}}')
+
+        initial_manifest_definition = {
+            "claim_generator": "python_test",
```
gpeacock: remove claim_generator from examples, it is obsolete now. (keep claim_generator_info!)

tests/test_unit_tests.py
```diff
@@ -2323,6 +2324,334 @@ def test_builder_state_with_invalid_native_pointer(self):
...
+            "claim_generator_info": [{
+                "name": "python_test",
+                "version": "0.0.1",
+            }],
+            # claim version 2 is the default
+            # "claim_version": 2,
+            "format": "image/jpeg",
+            "title": "Python Test Image V2",
+            "ingredients": [],
```
gpeacock: no need to add an empty field here

tests/test_unit_tests.py
```diff
@@ -2323,6 +2324,334 @@ def test_builder_state_with_invalid_native_pointer(self):
         with self.assertRaises(Error):
             builder.set_no_embed()
 
+    def test_builder_add_action_to_manifest_no_auto_add(self):
+        # For testing, remove auto-added actions
+        load_settings('{"builder":{"actions":{"auto_placed_action":{"enabled":false}}}}')
```
gpeacock: combine these into one load_settings call
{"builder": { "actions": {"auto_placed_action":{"enabled":false}, "auto_opened_action":{"enabled":false}, "auto_created_action":{"enabled":false} }}

gpeacock: I think we should get rid of these or at least not enable them by default. I'm moving toward using 
builder: intent instead.

src/c2pa/c2pa.py
```diff
@@ -2613,6 +2618,38 @@ def add_ingredient_from_file_path(
         except Exception as e:
             raise C2paError.Other(f"Could not add ingredient: {e}") from e
 
+    def add_action(self, action_json: str) -> None:
```
gpeacock: fyi, I think that for python we should allow either a string or a dict to pass in here, It's a lot more convenient to define the action as a dictionary and just pass it in instead of needing to stringify it first. i think builder allows that.

tests/test_unit_tests.py
tmathern: Let me know once we have the Builder::intent APIs, and I'll add support in Python!

src/c2pa/c2pa.py
tmathern: Sounds good. While at it, I make sure add_ingredient also supports dict.
Also, load_settings now handled both string and dict for Python.

src/c2pa/c2pa.py
```diff
@@ -667,20 +671,45 @@ def version() -> str:
...
 
 
+@overload
 def load_settings(settings: str, format: str = "json") -> None:
-    """Load C2PA settings from a string.
+    ...
+
+
+@overload
+def load_settings(settings: dict) -> None:
```
tmathern, line 680: Overload to handle settings coming in as dict

src/c2pa/c2pa.py
```diff
@@ -667,20 +671,45 @@ def version() -> str:
     return _convert_to_py_string(result)
 
 
+@overload
 def load_settings(settings: str, format: str = "json") -> None:
```
tmathern, line 675: Overload to handle settings coming in as string (mostly for API type checking)

src/c2pa/c2pa.py
tmathern: (+tests for those variations)

PR #189, opened 2025-10-24
feat: c2pa-rs and c2pa_c_ffi 0.68.0 update

tmathern:
- Bump c2pa-rs version to 0.68.0.
- Add a file with default trust settings (excerpt from the reference test settings file from c2pa-rs), dedicated for use in tests.
- Add some tests that load trust settings. (Older test are updated to ignore validation results or check that the result is Invalid, since without trust config most manifests will be Invalid. This also makes sure we have tests covering validation... indirectly). Tests which load trust settings run isolated in their own thread each.
- Add according notes to the example files.

Reference test settings, also used in c2pa-rs tests: https://github.com/contentauth/c2pa-rs/blob/main/sdk/tests/fixtures/test_settings.toml


tests/test_unit_tests_threaded.py
```diff
@@ -1429,10 +1429,6 @@ def sign_file(output_stream, manifest_def, thread_id):
             active_manifest1["title"],
             active_manifest2["title"])
 
-        # Verify both outputs have valid signatures
```
tmathern, line 1432: Skipping that validation for those tests.

tests/test_unit_tests_threaded.py
```diff
@@ -1511,12 +1507,6 @@ async def read_manifest():
                 self.assertTrue(author_found,
                                 "Author assertion not found in manifest")
 
-                # Verify no validation errors
```
tmathern, line 1514: Skipping that validation for those tests.

tests/trust_config_test_settings.toml
```diff
@@ -0,0 +1,239 @@
+# c2pa-rs Configuration File
```
gpeacock: It may be better to format this as JSON from the start in Python. I'm not sure anyone really wants to deal with toml in the Python world.

tests/test_unit_tests_threaded.py
gpeacock, line 1514: Were you running into errors here or are you just cleaning things up to not look for validation_status vs validation_results? Maybe we should add a validation_status or validation_state check here?

tmathern: Cleaning things up. Technically I would need to update the check here because it's a test without trust settings loaded. (I edited/indicated it in a lot of existing tests, but when I don't need it, especially in those threaded tests, I removed it).

tests/trust_config_test_settings.toml
tmathern: I can do both, I think, maybe? Since we're in a transition period? Basically the function I use in the tests already loads the toml into JSON.

tmathern: @gpeacock Done. I switched all tests to JSON except one.
I plan to keep that one test using TOML until we're officially and totally switched to JSON.

tests/test_unit_tests.py
```diff
@@ -38,9 +40,57 @@
 INGREDIENT_TEST_FILE = os.path.join(FIXTURES_DIR, INGREDIENT_TEST_FILE_NAME)
 ALTERNATIVE_INGREDIENT_TEST_FILE = os.path.join(FIXTURES_DIR, "cloud.jpg")
 
+
+def load_toml_test_settings_as_json():
```
gpeacock: Not sure we need to provide a sample for this now

tests/test_unit_tests.py
```diff
@@ -1134,9 +1388,70 @@ def test_sign_with_ps256_alg_2(self):
...
+                # Settings are thread-local, so they won't affect other tests
+                # And that is why we also run the test in its own thread, so tests are isolated
+                load_settings(settings_dict)
+
+                with open(os.path.join(self.data_dir, "ps256.pub"), "rb") as cert_file:
+                    certs = cert_file.read()
+                with open(os.path.join(self.data_dir, "ps256.pem"), "rb") as key_file:
+                    key = key_file.read()
+
+                signer_info = C2paSignerInfo(
```
gpeacock, line 1356: We can replace the signer setup with Signer.from_settings() soon

gpeacock (approved):

tests/test_unit_tests.py
tmathern: Ok, I will drop it. Especially since the JSON-only APIs are coming!

tests/test_unit_tests.py
tmathern: Yes, linking https://github.com/contentauth/c2pa-rs/pull/1533 for reference.

PR #207, opened 2025-12-30
Expose add example read.py script

Res260:
Add an example read.py script to show how to read a file's C2PA manifest and verify it.

### DIFF
--- a/src/c2pa/__init__.py
+++ b/src/c2pa/__init__.py
@@ -12,8 +12,8 @@
 # each license.
 
 try:
-    from importlib.metadata import version as get_version
-    __version__ = get_version("c2pa-python")
+    from importlib.metadata import version
+    __version__ = version("c2pa-python")
 except ImportError:  # pragma: no cover
     __version__ = "unknown"
 
@@ -26,13 +26,10 @@ from .c2pa import (
     Signer,
     Stream,
     sdk_version,
-    read_ingredient_file,
-    load_settings,
-    version
+    read_ingredient_file
 )  # NOQA
 
 # Re-export C2paError and its subclasses
-
 __all__ = [
     'Builder',
     'C2paError',
@@ -42,7 +39,5 @@ __all__ = [
     'Signer',
     'Stream',
     'sdk_version',
-    'version',
-    'read_ingredient_file',
-    'load_settings',
+    'read_ingredient_file'
 ]


examples/read.py
```diff
@@ -0,0 +1,43 @@
...
+if __name__ == '__main__':
+    if len(sys.argv) < 2:
+        print(f"Usage: python {sys.argv[0]} <image_path>")
+        sys.exit(1)
+
+    load_trust_anchors()
+
+    image_path = sys.argv[1]
+    try:
+        with c2pa.Reader(image_path) as reader:
```
tmathern: I get why you write it this way, but if you put trust settings loading into a function, you can also put the reading in a (very) small function too. So in main you only have the function calls.

ps: You should probably add some simple checks that the path is actually an image (it seems you are aiming to read only images, given the variable name).

src/c2pa/__init__.py
```diff
@@ -39,5 +41,7 @@
     'Signer',
     'Stream',
     'sdk_version',
-    'read_ingredient_file'
+    'read_ingredient_file',
```
tmathern: This is deprecated, please do not make that export more visible by adding it here. It wasn't made that visible on purpose, since deprecated. The function is still here for backwards compatibility purposes.

See https://github.com/contentauth/c2pa-python/blob/main/src/c2pa/c2pa.py#L770

src/c2pa/__init__.py
```diff
@@ -12,8 +12,8 @@
 # each license.
 
 try:
-    from importlib.metadata import version
-    __version__ = version("c2pa-python")
+    from importlib.metadata import version as get_version
+    __version__ = get_version("c2pa-python")
```
tmathern: Why this change? They do the same thing. Concerns on namespacing?

src/c2pa/__init__.py
```diff
@@ -39,5 +41,7 @@
     'Signer',
     'Stream',
     'sdk_version',
-    'read_ingredient_file'
+    'read_ingredient_file',
+    'load_settings',
+    'version'
```
tmathern: Sidenote: Could you move it next to sdk_version?
sdk_version is the version of the underlying native library.

I am also curious why you need this one exported?

README.md
```diff
@@ -34,6 +34,7 @@ import c2pa
 
 See the [`examples` directory](https://github.com/contentauth/c2pa-python/tree/main/examples) for some helpful examples:
 
+- `examples/read.py` shows how to read and verify an asset with a C2PA manifest.
```
tmathern, line 37: Since the example is not self contained, either pick an iamge from the test set to use, or describe how to use/call the script here.

tmathern:
fyi, there is also an example repo: https://github.com/contentauth/c2pa-python-example.
It is a more complete example than the examples in this repo.

Res260:
> fyi, there is also an example repo: https://github.com/contentauth/c2pa-python-example. It is a more complete example than the examples in this repo.

This is good to know and in my opinion should be advertized in this repo somehow (like in the README). I would not have found it if you didn't tell me!

README.md
Res260: Supported a default image if none is given

src/c2pa/__init__.py
Res260: I just wanted to avoid using `version` for two different purposes (one is for the python library version, the other is for the c2pa-rs version)

src/c2pa/__init__.py
Res260: For logging purposes. Like to use like this: `logger.info("Initializing C2PA reader using {c2pa.version()}")`. I can use `sdk_version()` but I liked that `version()` gives more details (`c2pa-c-ffi/0.73.1 c2pa-rs/0.73.1` vs `0.73.1`). Let me know

examples/read.py
Res260: extracted function, renamed variable from image_path to media_path

src/c2pa/__init__.py
tmathern: I see! This is mostly for historical reasons that we have 2.
c2pa-c-ffi and c2pa-rs are in lockstep release (we now always release c2pa-c-ffi when we release c2pa-rs, because they share a build system to create the native libraries), so sdk_version() give you their shared version number (so, 0.73.1 for instance). If I clarify that in the docs and comments, do you still wish to have both version and sdk_version then?

examples/read.py
tmathern: Thanks

src/c2pa/__init__.py
Res260: I thought it was simply forgotten, but it's really not a dealbreaker to not have it

tmathern: I prefer we'd not add it then. I will clarify the docs (hopefully this week) regarding the version and how it works then.
Once the "new" version is removed, this PR would be good to go I think (I'll have the tests run on the change before that).

tmathern: Like this: https://github.com/contentauth/c2pa-python/pull/211
(And then we'd merge this PR to keep your nice examples!).

Res260: Alright I removed the changes to `__init__.py`, should be good now

PR #212, opened 2026-01-06
fix: Add a try_create method on Reader

tmathern:
### Add Reader factory
Adds a `try_create` method on Reader class (as class method) that acts as a factory method for Reader objects:
- if there is JUMBF (c2pa/manifest) data to read, from_asset will return a Reader instance
- if there is no JUMBF (c2pa/manifest) data to read, returns None instead of throwing

### Fix exception classes hierarchy
- Fixes exception hierarchies to be able to leverage typed exceptions better
- Still keeps backwards compatibility with previous exception handling

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ See the [`examples` directory](https://github.com/contentauth/c2pa-python/tree/m
 
 Documentation is published at [github.io/c2pa-python/api/c2pa](https://contentauth.github.io/c2pa-python/api/c2pa/index.html).
 
-To build documentation locally, refer to [the section in Contributing to the project](https://github.com/contentauth/c2pa-python/blob/main/docs/project-contributions.md#api-reference-documentation).
+To build documentation locally, refer to [this section in Contributing to the project](https://github.com/contentauth/c2pa-python/blob/main/docs/project-contributions.md#api-reference-documentation).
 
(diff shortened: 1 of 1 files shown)


src/c2pa/c2pa.py
```diff
@@ -537,71 +533,114 @@ def _setup_function(func, argtypes, restype=None):
 
 
 class C2paError(Exception):
-    """Exception raised for C2PA errors."""
+    """Exception raised for C2PA errors.
```
tmathern, line 536: Fixing error class hierarchy to leverage typed errors better, especially in the new Reader factory method

src/c2pa/c2pa.py
```diff
@@ -537,71 +533,118 @@ def _setup_function(func, argtypes, restype=None):
...
+
+    This is the base class for all C2PA exceptions. Catching C2paError will
+    catch all typed C2PA exceptions (e.g., C2paError.ManifestNotFound).
+    """
 
     def __init__(self, message: str = ""):
         self.message = message
         super().__init__(message)
 
-    class Assertion(Exception):
```
tmathern, line 546: Specific exceptions got un-nested from here so they can inherit from the base C2pa exception and get the C2paException type.

PR #213, opened 2026-01-07
fix: Improve typings for errors

tmathern:
- Fixes exception hierarchies to be able to leverage typed exceptions better
- Still keeps backwards compatibility with previous exception handling

## Checklist

### DIFF
--- a/src/c2pa/c2pa.py
+++ b/src/c2pa/c2pa.py
@@ -1452,39 +1452,6 @@ class Reader:
     }
 
     @classmethod
-    def try_create(cls,
-                   format_or_path: Union[str, Path],
-                   stream: Optional[Any] = None,
-                   manifest_data: Optional[Any] = None) -> Optional["Reader"]:
-        """This is a factory method to create a new Reader from an asset,
-        returning None if no manifest/c2pa data/JUMBF data could be read
-        (instead of raising a ManifestNotFound: no JUMBF data found exception).
-
-        Returns None instead of raising C2paError.ManifestNotFound if no
-        C2PA manifest data is found in the asset. This is useful when you
-        want to check if an asset contains C2PA data without handling
-        exceptions for the expected case of no manifest.
-
-        Args:
-            format_or_path: The format or path to read from
-            stream: Optional stream to read from (Python stream-like object)
-            manifest_data: Optional manifest data in bytes
-
-        Returns:
-            Reader instance if the asset contains C2PA data,
-            None if no manifest found (ManifestNotFound: no JUMBF data found)
-
-        Raises:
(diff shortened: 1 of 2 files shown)


src/c2pa/c2pa.py
```diff
@@ -537,71 +533,118 @@ def _setup_function(func, argtypes, restype=None):
 
 
 class C2paError(Exception):
-    """Exception raised for C2PA errors."""
+    """Exception raised for C2PA errors.
```
tmathern, line 536: Fixing error class hierarchy to leverage typed errors better, especially in the new Reader factory method

src/c2pa/c2pa.py
```diff
@@ -537,71 +533,118 @@ def _setup_function(func, argtypes, restype=None):
...
+
+    This is the base class for all C2PA exceptions. Catching C2paError will
+    catch all typed C2PA exceptions (e.g., C2paError.ManifestNotFound).
+    """
 
     def __init__(self, message: str = ""):
         self.message = message
         super().__init__(message)
 
-    class Assertion(Exception):
```
tmathern, line 546: Specific exceptions (all the types nested in the parent class) got un-nested from here so they can inherit from the base C2pa exception and get the C2paException type.

src/c2pa/c2pa.py
```diff
@@ -537,71 +533,118 @@ def _setup_function(func, argtypes, restype=None):
...
-        """Exception raised for signature errors."""
-        pass
 
-    class Verify(Exception):
-        """Exception raised for verification errors."""
-        pass
+# Attach exception subclasses to C2paError for backward compatibility
+# Preserves behavior for exception catching like except C2paError.ManifestNotFound,
+# also reduces imports (think of it as an alias of sorts)
+C2paError.Assertion = _C2paAssertion
```
tmathern, line 633: An we still can use the parent type for general error catching!

src/c2pa/c2pa.py
```diff
@@ -656,60 +699,106 @@ def _convert_to_py_string(value) -> str:
...
+
+    Raises:
+        C2paError subclass: The appropriate typed exception based on error_str
+    """
+    # Error format from native library is "ErrorType: message" or "ErrorType message"
+    # Try splitting on ": " first (colon-space), then fall back to space only
+    if ': ' in error_str:
+        parts = error_str.split(': ', 1)
+    else:
+        parts = error_str.split(' ', 1)
```
gpeacock, line 720: If the c api is inconsistent here, we should add an issue to fix it so that we always use the colon

gpeacock (approved):

Issue #265, opened 2026-05-22
Feature Request: Add Windows ARM64 (aarch64-pc-windows-msvc) support


JDRkow:
## Summary

Currently `c2pa-python` does not provide prebuilt wheels for Windows ARM64 (`aarch64-pc-windows-msvc`).
Installing via `pip install c2pa-python` on a `windows-11-arm` GitHub Actions runner fails because
`setup.py` has no entry for `aarch64-pc-windows-msvc` in `PLATFORM_FOLDERS`.

## Request

1. Add `aarch64-pc-windows-msvc` to `PLATFORM_FOLDERS` in `setup.py`
2. Publish a prebuilt wheel `c2pa_python-*-py3-none-win_arm64.whl` to PyPI

tmathern:
https://github.com/contentauth/c2pa-python/pull/267 has been merged. v0.33.0 will be the first released version with that winARM support.

PR #267, opened 2026-05-26
feat: Add support for winARM in build and release pipeline

tmathern:
Add support to run tsts on winARM, and publish winARM wheels.
Successful publish on PyPI test: https://test.pypi.org/manage/project/c2pa-python/release/0.33.0/

For https://github.com/contentauth/c2pa-python/issues/265

## Checklist


.github/workflows/build-wheel.yml
```diff
@@ -88,7 +88,17 @@ jobs:
           echo "Verifying wheels using twine check"
           twine check dist/*
 
-      - name: Build Windows wheel (x64)
+      - name: Install ARM64 OpenSSL via vcpkg (Windows ARM64)
```
tmathern, line 91: Needed so this can even think of building...

.github/workflows/build.yml
```diff
@@ -128,7 +128,16 @@ jobs:
...
+    runs-on: ${{ matrix.runs-on }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - runs-on: windows-latest
+            python-version: "3.10"
+          - runs-on: windows-11-arm
+            python-version: "3.11"  # win-arm runner needs 3.11 atleast
```
tmathern, line 140: WinARM runner doesn't have Python 3.11, and I don't want to hand-compile it (it's slow and more involved). Using Python 3.11 for this one should be fine.

.github/workflows/build.yml
```diff
@@ -137,9 +146,20 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: ${{ matrix.python-version }}
           cache: "pip"
 
+      - name: Install ARM64 OpenSSL via vcpkg (Windows ARM64)
```
tmathern, line 152: Still need that one to even think of running tests... (cryptography crate needs it too, and we use it in tests also)

.github/workflows/build.yml
```diff
@@ -294,10 +318,17 @@ jobs:
...
+      fail-fast: false
       matrix:
-        target: [x64]
+        include:
+          - target: x64
+            runs-on: windows-latest
+            python-version: "3.10"
+          - target: arm64
+            runs-on: windows-11-arm
+            python-version: "3.11"  # win-arm runner needs 3.11 atleast
```
tmathern, line 331: No Python 3.10 OOTB on winARM runner. Python 3.11 should be fine here.

src/c2pa/lib.py
```diff
@@ -54,6 +55,8 @@ def get_platform_identifier() -> str:
         else:
             return "universal-apple-darwin"
     elif system == "windows":
+        if _get_architecture() in [CPUArchitecture.ARM64.value, CPUArchitecture.AARCH64.value]:
```
tmathern, line 58: Making sure to load the proper library.

src/c2pa/lib.py
```diff
@@ -81,7 +84,7 @@ def _get_architecture() -> str:
     elif sys.platform == "win32":
         # win32 will cover all Windows versions
         # (the 32 is a historical quirk)
-        return platform.machine()
+        return platform.machine().lower()
```
tmathern, line 87: Different OSes have different opinion on casing, so we'll make them agree.

.github/workflows/build.yml
ok-nick: Why not use 3.11 for both to keep it consistent?

.github/workflows/build.yml
ok-nick: What is the issue here? I'm surprised we need this because the Rust library should already be statically linking to openssl.

.github/workflows/build.yml
tmathern: Technically we support back 3.10. I want to keep most of the coverage on 3.10 due to that.

.github/workflows/build.yml
tmathern: The cryptography crate doesn't find its right expected version or OpenSSL. This makes tests crash.

ok-nick: Is that a dependency of c2pa? Do you know which one specifically so I can take a deeper look?

tmathern: No, I am talking of the Python cryptography package

tmathern: https://github.com/contentauth/c2pa-python/blob/main/requirements-dev.txt#L19

tmathern: (So I typo'ed. It's not crate, it's package. Python only here at work. crypto package needs openssl for the tests. Note it's a test/dev dependency and not a package dependency, so it isn't included in the wheel).

ok-nick: Oh ok. If you can provide more details about the failure that would be great for future readers.

ok-nick (approved):

.github/workflows/build.yml
tmathern: So the symptom is the following:
```
from cryptography.hazmat.primitives import hashes, serialization
  File "C:\hostedtoolcache\windows\Python\3.11.9\arm64\Lib\site-packages\cryptography\hazmat\primitives\hashes.py", line 9, in <module>
    from cryptography.hazmat.bindings._rust import openssl as rust_openssl
```
This is indicative of the missing expected "proper openssl". Note how it comes from the `cryptography` package.

PR #269, opened 2026-05-29
fix: Close Builder after signing

tmathern:
- Close Builder to force resources clean up
- A Builder should not be reused for multiple signing anyway

## Checklist

### DIFF
--- a/src/c2pa/c2pa.py
+++ b/src/c2pa/c2pa.py
@@ -3529,15 +3529,9 @@ class Builder(ManagedResource):
                     dest_stream._stream,
                     ctypes.byref(manifest_bytes_ptr),
                 )
-            # c2pa_builder_sign / c2pa_builder_sign_context BORROW the builder
-            # (&mut self on the Rust side); they do NOT take ownership. The
-            # native Builder is still owned by us and must be freed. Marking it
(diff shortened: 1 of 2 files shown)


src/c2pa/c2pa.py
```diff
@@ -3529,10 +3529,12 @@ def _sign_internal(
...
                 )
-            # Builder pointer consumed by Rust FFI at this point
-            self._mark_consumed()
+            # Sign borrows the Builder without taking ownership.
+            # Closing here ensures resources clean up,
+            # and single use/single sign done by a Builder.
+            self.close()
         except Exception as e:
-            self._mark_consumed()
+            self.close()
```
ok-nick, line 3539: Can we make sure this is documented in the function signatures just so it's not a surprise to anyone?
(not sure if that's done already or is any different from what we had previously)

ok-nick (approved):
Approving given the previous comment is resolved.

src/c2pa/c2pa.py
tmathern: (It is documented in the docs, I can add it to the signature too. It's a change I missed to do as we recently updated mutation rules in Rust).

tmathern: Updated: Added comment in all function paths that hit sign! (So it's very clear)

PR #273, opened 2026-06-03
fix: Add bindings for new archive APIs

tmathern:
- Sister PR of https://github.com/contentauth/c2pa-cpp/pull/223

## Checklist


docs/selective-manifests.md
```diff
@@ -494,6 +494,17 @@ An **ingredient archive** contains the manifest store from an asset that was add
...
 The key difference: a builder archive is a work-in-progress (unsigned). An ingredient archive carries the provenance history of a source asset for reuse as an ingredient in other working stores.
 
+### Producing an ingredient archive
+
+The SDK supports two approaches for producing an ingredient archive. They share the same `.c2pa` binary format and are interchangeable from the consumer side.
+
+| Approach | Entry point | Status |
+| --- | --- | --- |
+| Dedicated ingredient archive APIs | `add_ingredient` then `write_ingredient_archive(id, stream)` | **Recommended** |
+| Read-filter-rebuild pattern | `Builder` + `add_ingredient` + `to_archive`, then `Reader` + manual JSON | **Older pattern** |
```
ok-nick: This is a little bit unclear to me. The old method only preserved the parent ingredient, but the reader and manual JSON aren't involved there.

docs/selective-manifests.md
```diff
@@ -557,6 +568,132 @@ with Reader("application/c2pa", archive_stream, context=ctx) as reader:
...
+final_builder.add_ingredient_from_archive(archive_b)
+with open("source.jpg", "rb") as src, open("output.jpg", "w+b") as dst:
+    final_builder.sign(signer, "image/jpeg", src, dst)
+```
+
+Action linking also changes between the two approaches. Legacy catalog code linked ingredients via `label` set on the signing builder's `add_ingredient` JSON; `instance_id` was not accepted. The dedicated archive API accepts the archive key passed to `write_ingredient_archive`, which can be either `label` or `instance_id`. See [Lookup keys and action linking](working-stores.md#lookup-keys-and-action-linking).
+
+#### Choosing between approaches
+
+The legacy read-filter-rebuild APIs fit when the catalog already exists as one multi-ingredient builder archive and the consumer wants a subset of it. The dedicated ingredient archive APIs fit when ingredients are produced and consumed independently: each archive holds exactly one ingredient, and the call sites stay short. Both produce the same signed output.
```
ok-nick: For ingredient archives, doesn't the legacy only support one parent ingredient?

tmathern: No, we switched to support component to, it became agnostic of the relationship.
The one ingredient only is true, which is why we've asked in the docs to only have one ingredient in those archives!

docs/selective-manifests.md
tmathern: only preserved the parent ingredient <- no, it became relationship agnostic! The reader and manual JSON were mentioned because this is the selective manifest docs explaining how to cherry pick.

ok-nick (approved):

PR #277, opened 2026-06-05
chore: Enable to build c2pa-python with a local c2pa-rs build

tmathern:
Enables debug builds based on checked out c2pa-rs/c2pa-c-ffi

## Checklist


scripts/build_local_artifacts.py
```diff
@@ -0,0 +1,268 @@
...
+import shutil
+import argparse
+import platform
+import subprocess
+from pathlib import Path
+
+# The crate in c2pa-rs that produces the native library.
+FFI_PACKAGE = "c2pa-c-ffi"
+# Extra c2pa-c-ffi features to enable on top of the crate defaults
+FFI_FEATURES = "file_io"
```
ok-nick, line 41: Should this be configurable?

ok-nick (approved):

scripts/build_local_artifacts.py
```diff
@@ -0,0 +1,268 @@
...
+        sys.exit(1)
+    except subprocess.CalledProcessError as e:
+        print(f"Error: cargo clean failed (exit code {e.returncode}).")
+        sys.exit(e.returncode)
+
+
+def run_cargo(c2pa_rs_path, extra_args=None):
+    """Run a release build of the FFI crate in the c2pa-rs checkout."""
+    cmd = ["cargo", "build", "--release", "-p", FFI_PACKAGE,
+           "--features", FFI_FEATURES]
```
ok-nick: Specifying the debug build rather than release would be nice as well.

ok-nick (commented):
How does this script differ from the normal build workflow? It seems like it should be doing a lot of the the same work?

tmathern:
> How does this script differ from the normal build workflow? It seems like it should be doing a lot of the the same work?

The assumptions differ. This one *assumes* you have a local c2pa-rs checkout next to this repo. The normal workflow takes/downloads a prebuilt released library (and does no rebuild). So it also asks you to have a local cargo setup (which you may not have if developing only Python code).

scripts/build_local_artifacts.py
tmathern: Technically, no, since it should match the config of the pre-built library.

tmathern: Or rather, not yet until we decide (if ever) to add features.

scripts/build_local_artifacts.py
tmathern: Done

PR #282, opened 2026-06-09
fix: API updates to remove long-deprecated APIs

tmathern:
Follows suit to https://github.com/contentauth/c2pa-rs/pull/2206

APIs removed: `read_file`, `read_ingredient_file`, `sign_file`.

## Checklist


src/c2pa/c2pa.py
```diff
@@ -889,33 +886,49 @@ def __init__(self):
...
+# Wrap a raw (address, length) native region in a writable memoryview
+_PyMemoryView_FromMemory = ctypes.pythonapi.PyMemoryView_FromMemory
+_PyMemoryView_FromMemory.restype = ctypes.py_object
+_PyMemoryView_FromMemory.argtypes = (
+    ctypes.c_void_p, ctypes.c_ssize_t, ctypes.c_int)
+_PyBUF_WRITE = 0x200
+
+
+def _writable_memoryview(address, length):
+    return _PyMemoryView_FromMemory(address, length, _PyBUF_WRITE)
```
ok-nick: I'm reading that these APIs only work with CPython, not other runtimes like PyPy. Is that a concern?

src/c2pa/c2pa.py
```diff
@@ -1827,10 +1637,19 @@ def read_callback(ctx, data, length):
...
-                    return n if n else 0
+                    buf = _writable_memoryview(
+                        ctypes.addressof(data.contents), length)
+                    try:
+                        n = readinto(buf)
+                    finally:
+                        # Invalidate the view:
+                        # The native buffer is only valid for
+                        # the duration of the callback...
+                        buf.release()
```
ok-nick: Just want to be sure, this doesn't release the buffer passed in as a parameter?

src/c2pa/c2pa.py
```diff
@@ -1827,10 +1637,19 @@ def read_callback(ctx, data, length):
...
+                    if not n:
+                        return 0
+                    # Never report more than the buffer can hold
+                    return min(n, length)
```
ok-nick: I think this should be an error if n>length. If that happens something went wrong somewhere and ignoring it can make the problem worse?

src/c2pa/c2pa.py
tmathern: I don't think so yet. Technically we only though of CPython (which I think happens to be the very most common), we can adapt if the issue is raised.

src/c2pa/c2pa.py
tmathern: Correct, this is the release of the wrapper we create a few lines above!

src/c2pa/c2pa.py
tmathern: A check+wrapper with fallback is quite easy to add though. Let me throw it in.

src/c2pa/c2pa.py
tmathern: Done

ok-nick (approved):

PR #286, opened 2026-06-16
fix: Remove some other potential gc object reference loops

tmathern:
Remove some other potential gc object reference loops to relieve gc pressure.

## Checklist

### DIFF
--- a/src/c2pa/c2pa.py
+++ b/src/c2pa/c2pa.py
@@ -18,6 +18,7 @@ import logging
 import sys
 import os
 import warnings
+import weakref
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Optional, Union, Callable, Any, overload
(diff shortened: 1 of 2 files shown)


src/c2pa/c2pa.py
ok-nick, line 1: Isn't it entirely possible that Python doesn't hold a reference to the stream but the Rust still operates on it (perhaps writing to a file is still ongoing). We wouldn't want it to stop in the middle of that operation?

tmathern: We hold the streams in local varibales during calls (that "fortifies" weak refs, they won't just disappear mid-execution => they are referenced and gc won't pick them up. That refcount takes even if the GIL is released aka during Rust execution, so we are good!). Finally cleans up things only after returns.

ok-nick (approved):

PR #287, opened 2026-06-22
fix: Refactor checks for manifest data when binding for `c2pa_reader_with_manifest_data_and_stream` is used

tmathern:
- Refactor manifest check on top of constructor for binding of `c2pa_reader_with_manifest_data_and_stream` (binding was already there, this improve on testing and does a refactor)
- Refactor benchmark checks to only fail on leaks detection
- Refactor tests for sidecars in their own test class
- Added benchmark for sidecar file reading
Screenshot 2026-06-21 at 8 44 16 PM

## Checklist

### DIFF
--- a/tests/perf/baseline.json
+++ b/tests/perf/baseline.json
@@ -186,5 +186,10 @@
     "peak_bytes": 3382176,
     "leaked_bytes": 3210242,
     "total_allocations": 111383
+  },
+  "reader_manifest_data_context": {
+    "peak_bytes": 7575548,
+    "leaked_bytes": 3399851,
+    "total_allocations": 1414270
   }
 }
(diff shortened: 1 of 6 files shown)


tests/perf/baseline.json
```diff
@@ -186,5 +186,10 @@
     "peak_bytes": 3382176,
     "leaked_bytes": 3210242,
     "total_allocations": 111383
+  },
+  "reader_manifest_data_context": {
```
gpeacock, line 190: quite a leap here in peak_bytes and total_allocations, do you know why?

tests/perf/run_profile.py
```diff
@@ -242,16 +251,25 @@ def main() -> None:
 
             if baseline and name in baseline:
                 b = baseline[name]
+                # Only leaked_bytes gates the run. It is the leak signal and is
```
gpeacock, line 254: this makes we wonder if we should expose an API for c_ffi leaked bytes. But I'm not sure exactly how that would work.

gpeacock (approved):
Were we having leaks here? How did you know about them?

tmathern:
> Were we having leaks here? How did you know about them?

Yes, we did have one leak recently in c2pa-python in general. A Builder handle which was not released when it should. Accumulated 49 bytes over time on each Builder instance.

So I've been adding scenarios since then when I make change to make sure heap usage stays flat.

PR #294, opened 2026-07-16
fix: Centralize ManagedResources lifecycle handling

tmathern:
Centralizes ManagedResources lifecycle plumbing (the managed resources holding a native handle) natively and handle their lifecycle. This change introduces a set of new (private) functions on the ManagedResource class for lifecycle management. Some notable additions are:

- `_activate(handle)`: take ownership of a validated non-null pointer and mark the resource active. Rejects null and double-activation, so a handle can't be activated twice and a closed resource can't be reopened.
- `_swap_handle(new_handle)`: replace the handle after an FFI call consumed the old one and returned a replacement.
- `_consume_*` (family of functions): runs an FFI call that consumes this handle and manages the handle lifecycle as part of the object's lifecycle.
- `_wrap_native_handle(handle)`: build an instance around an already-owned pointer, bypassing __init__.
- `_init_attrs()`: subclass hook for attribute defaults, called by both __init__ and _wrap_native_handle so an instance built around an existing handle can't be missing state the class reads.

Those are all private APIs for native handling (hence underscore in function name, using and calling them must be very explicit).

Note on fork safety: the owner PID is still stamped on all paths, and the native free is skipped in a foreign process on all teardown paths.

Other changes:
- Improve error handling by removing function that did not clear previous error (C FFI doesn't clear/do a take on the error either).
- Switch examples to print crJson (instead of "just JSON").

## Checklist


tests/perf/entrypoint.sh
```diff
@@ -16,8 +16,14 @@ else
     PLATFORM="x86_64-unknown-linux-gnu"
 fi
 
-echo "Downloading c2pa native lib: $C2PA_VERSION / $PLATFORM"
-C2PA_LIBS_PLATFORM=$PLATFORM python scripts/download_artifacts.py "$C2PA_VERSION"
+# Skip the GitHub API round-trip when the lib is already on disk
+# Set C2PA_FORCE_DOWNLOAD=1 to override.
+if [ -z "$C2PA_FORCE_DOWNLOAD" ] && [ -f "artifacts/$PLATFORM/libc2pa_c.so" ]; then
```
tmathern, line 21: If lib is available locally already, bypass download on rerun (avoid hitting Github rate limits, potentially, when working on test changes that need rebuild)

docs/native-resources-management.md
```diff
@@ -105,10 +109,10 @@ The native Rust library exposes a single C FFI function, `c2pa_free`, that deall
 ```python
 @staticmethod
 def _free_native_ptr(ptr):
-    _lib.c2pa_free(ctypes.cast(ptr, ctypes.c_void_p))
+    _lib.c2pa_free(ptr)
 ```
 
-All native pointers are freed through this single path, regardless of which constructor created them (`c2pa_reader_from_stream`, `c2pa_builder_from_json`, `c2pa_signer_from_info`, etc.). The `ctypes.cast` to `c_void_p` is needed because the C function accepts a generic void pointer regardless of the original type.
+All native pointers are freed through this single path, regardless of which constructor created them (`c2pa_reader_from_stream`, `c2pa_builder_from_json`, `c2pa_signer_from_info`, etc.). No explicit `ctypes.cast` is needed: `c2pa_free`'s declared argtype is `c_void_p`, so ctypes converts any pointer instance on the way in. Casting explicitly with `ctypes.cast(ptr, c_void_p)` performs the same conversion but leaves a reference cycle behind on every call, which creates additional load on the (Python) garbage collector.
```
ale-adobe, line 154: 🎉 So much cleaner!

tests/perf/entrypoint.sh
```diff
@@ -16,8 +16,14 @@ else
...
+    echo "Using cached c2pa native lib: artifacts/$PLATFORM (set C2PA_FORCE_DOWNLOAD=1 to re-download)"
```
ale-adobe, line 22: Would it be worth adding a bit of logic here to check the version of the lib that's already on disk vs. what's listed in `c2pa-native-version.txt` and emitting a warning message to notify the developer? We don't have to force a download, but could be helpful to include the version numbers in the message for visibility.

src/c2pa/c2pa.py
```diff
@@ -274,36 +306,261 @@ def _release(self):
...
+        except Exception:
+            logger.error(
+                "Failed to release %s resources",
+                type(self).__name__,
+                exc_info=True,
+            )
+
+    def _teardown(self, free_handle: bool):
+        """Close the object: run _release, optionally free the handle, null it.
+        free_handle=False (consumed) frees nothing, the new owner needs to free.
```
ale-adobe, line 323: The wording here confuses me a bit. What does the (consumed) mean? My understanding is just that if `free_handle` is set as `False`, then the function doesn't free the handle, and it's the owner's responsibility to free it.

src/c2pa/c2pa.py
```diff
@@ -274,36 +306,261 @@ def _release(self):
...
+                             type(self).__name__, exc_info=True)
+
+    def _release_handle(self):
+        """Free this handle, then close the object. Used only where ownership is
+        unknown (a guarded free is a real free if ours, a no-op if not).
+        """
+        if self._lifecycle_state != LifecycleState.ACTIVE:
+            self._handle = None
+            self._lifecycle_state = LifecycleState.CLOSED
+            return
```
ale-adobe, line 348: This is just clean-up to make sure everything is in a sane state, right?

src/c2pa/c2pa.py
```diff
@@ -274,36 +306,261 @@ def _release(self):
...
+
+        self._handle = handle
+        self._lifecycle_state = LifecycleState.ACTIVE
+
+    def _create_and_activate(self, ffi_call, error_message, *,
+                             check=lambda r: not r):
+        """Obtain a fresh native pointer, validate it, and take ownership.
+        On any failure before ownership transfers, the pointer is freed
+        and the error re-raised.
+        """
```
ale-adobe, line 388: Can we add descriptions for the args and error behavior in the docstring?

src/c2pa/c2pa.py
```diff
@@ -1416,6 +1661,18 @@ class Context(ManagedResource, ContextProvider):
...
 
+    class _NativeBuilder(ManagedResource):
+        """Short-lived wrapper so the native context builder rides the normal
+        lifecycle: any failure inside its `with` block frees it via close()
+        unless a consuming call already took it.
+        """
+
+        def __init__(self):
+            super().__init__()
+            ptr = _lib.c2pa_context_builder_new()
```
ale-adobe: Could/should this use the new `_create_and_activate` helper? Looks equivalent, except for the try-except, although I don't think the call to `_lib.c2pa_context_builder_new()` could fail here anyways.

src/c2pa/c2pa.py
```diff
@@ -1433,72 +1690,46 @@ def __init__(
...
         if settings is None and signer is None:
             # Simple default context
-            ptr = _lib.c2pa_context_new()
+            context_ptr = _lib.c2pa_context_new()
             _check_ffi_operation_result(
-                ptr, "Failed to create Context"
+                context_ptr, "Failed to create Context"
             )
-            self._handle = ptr
+            self._activate(context_ptr)
```
ale-adobe: Same question here!

ale-adobe (approved):
LGTM, all comments and thoughts are non-blocking.

tests/perf/entrypoint.sh
tmathern: Debatable, since this is the memory perf bench. It's already listed in the baseline JSON (in the case this is used to regenerate a baseline), and you should see it on the build too. Also when there are changes, and you didn't update, the tests (unit) also fail due to the version number check.
So something should already fail pretty loudly if tests are run before benchmarks.

tmathern: (That one here: https://github.com/contentauth/c2pa-python/blob/main/tests/test_unit_tests.py#L69-L81
It checks you got the correct version of the native lib. I would not repeat it in the benchmarks, since I expect unit tests run as smoke test before the long running benchmarks. On CI/CD benchmarks are always a clean build).

src/c2pa/c2pa.py
tmathern: Correct! So the lifecycle is uniform.

src/c2pa/c2pa.py
tmathern: Will be in next pushed changeset

src/c2pa/c2pa.py
tmathern: Noted, one more refactoring coming.

src/c2pa/c2pa.py
tmathern: Noted, one more refactoring coming.

src/c2pa/c2pa.py
tmathern: Consumed (in general in this codebase) means the pointer is not yours to hold (manage) anymore, the native library took ownership. I will add a section in the docs on terms and vocabulary (in regard to this lifecycle).

examples/sign_info.py
```diff
@@ -40,7 +40,7 @@
 print("\nReading existing C2PA metadata:")
 with open(fixtures_dir + "C.jpg", "rb") as file:
     with c2pa.Reader("image/jpeg", file) as reader:
-        print(reader.json())
+        print(reader.crjson())
```
gpeacock: Are you recommending that python users always use crJson? It was designed for conformance testing, we haven't really figured out if / how it works with as a reader api for getting resources and such.

gpeacock (approved):
It feels like a failure of the c_ffi_bindings that it is this complex to track and handle things. I'd like to track what we can do to make this less of a burden on the bindings.
1) a better/ thread safe and atomic way to return error strings
2) consider using unique identifiers instead of actual reusable memory adresses.

examples/sign_info.py
tmathern: I thought the goal was to switch everyone to crJson?

PR #299, opened 2026-07-22
feat: Enable Reader to be more pass-through to guess mimetype

tmathern:
- Allow Reader to pass through an empty mimetype string to the native library, so the native lib attempts to guess the type of file.
- Still request the mimetype/format on Builder when signing.

Stream-only overloads for Reader will be in https://github.com/contentauth/c2pa-python/pull/301.

## Checklist


src/c2pa/c2pa.py
```diff
@@ -2394,15 +2431,16 @@ def _get_cached_manifest_data(self) -> Optional[dict]:
 
         return self._manifest_data_cache
 
-    def with_fragment(self, format: str, stream,
+    def with_fragment(self, format: Optional[str], stream,
```
ale-adobe, line 2434: Can we add a new unit test to cover the case(s) when empty string and/or `None` are provided for `format`?

tests/test_unit_tests.py
```diff
@@ -326,12 +420,193 @@ def test_reader_detects_unsupported_mimetype_on_file(self):
...
+        # Only the final suffix determines the type.
+        self.assertEqual(
+            _get_mime_type_from_path("photo.final.jpg"), "image/jpeg")
+
+    def test_get_mime_type_from_path_compound_extension(self):
+        # mimetypes treats ".gz" as an encoding, not a type suffix,
+        # the type is derived from ".tar"
+        result = _get_mime_type_from_path("archive.tar.gz")
+        self.assertIsInstance(result, str)
+        self.assertNotIn(result, ("image/jpeg", "image/dng"))
```
ale-adobe, line 583: Is `mimetypes` able to guess `.tar`? Or are we unable to test any further than this?

src/c2pa/c2pa.py
```diff
@@ -1969,27 +1973,51 @@ def _get_supported_mime_types(ffi_func, cache):
...
-        C2paError.Encoding: If the string contains invalid UTF-8 characters
+        C2paError.NotSupported: If a given format is unsupported, or if the
+            format is missing and ``allow_autodetect`` is False.
+        C2paError.Encoding: If the format contains invalid UTF-8 characters.
     """
-    if format_str.lower() not in supported_types:
+    key = (format_str or "").strip()
+    if not key:
+        if allow_autodetect:
+            return b""
```
ok-nick, line 2011: Can we make the return type optional instead of returning empty bytes?

src/c2pa/c2pa.py
```diff
@@ -1196,16 +1196,21 @@ def load_settings(settings: Union[str, dict], format: str = "json") -> None:
 
 
 def _get_mime_type_from_path(path: Union[str, Path]) -> str:
```
ok-nick, line 1198: Same here, can we return an optional. I think the caller should check if it's `None` to pass it to the SDK as an empty string.

src/c2pa/c2pa.py
```diff
@@ -1969,27 +1973,51 @@ def _get_supported_mime_types(ffi_func, cache):
...
 
     Raises:
-        C2paError.NotSupported: If the format is not supported
-        C2paError.Encoding: If the string contains invalid UTF-8 characters
+        C2paError.NotSupported: If a given format is unsupported, or if the
+            format is missing and ``allow_autodetect`` is False.
+        C2paError.Encoding: If the format contains invalid UTF-8 characters.
     """
-    if format_str.lower() not in supported_types:
+    key = (format_str or "").strip()
```
ok-nick, line 2008: IMO we should pass as is, the Rust SDK can determine if there's a problem.

src/c2pa/c2pa.py
tmathern: Yes and no. Yes, but not in this PR.
No: Not here because the native lib takes this as flag to guess.
Yes: I have the overloads in https://github.com/contentauth/c2pa-python/pull/301 for that. I split them out because I wanted to give them extra care.

src/c2pa/c2pa.py
tmathern: Same note as other comment - in other PR. I wanted to give them extra care.

src/c2pa/c2pa.py
tmathern: Fail fast...

ok-nick: The Rust SDK should fail fast, I think we should avoid adding extra overhead in the bindings.

src/c2pa/c2pa.py
ok-nick: I prefer to return an optional so that it's clear what this function does in the type system. Not a blocker.

ok-nick (approved):

src/c2pa/c2pa.py
```diff
@@ -2113,37 +2142,46 @@ def try_create(
...
         context: Optional['ContextProvider'] = None,
     ):
         """Create a new Reader.
 
+        The format is optional: pass None, an empty string, or read an
+        extensionless file by path to let the native library detect the
+        format. Detection is best effort, so pass the format when you know it.
+        A recognized but wrong format/extension is also corrected from the bytes
+        when reading (so reading doesn't fail on wrong file extension).
+        If the bytes are not a recognized asset, a C2paError is raised.
```
gpeacock: The internal code can do more when we get a format to work with than when we get nothing.
So we are losing some functionality this way.

If the sdk gets nothing, we try to detect the container type and make our best guess at a format. If we don't find a match I think an error is raised.

If we get a format, we compare the detected container type to the format given, if it is the same container, we use the given format type. If they aren't the same container, we ignore the format type and return a best guess for the correct container. If we can't find a container, we can use the format and presume it will be treated as a sidecar.

gpeacock: I'm not sure how the python support corresponds to this. We may need to export some other API to help.

gpeacock (commented):
I'm not sure if this captures the same thing that the sdk does, or if it should.

src/c2pa/c2pa.py
tmathern: `If we don't find a match I think an error is raised.`: Yes there is an error raised in this case!

ale-adobe (approved):

src/c2pa/c2pa.py
tmathern: Ok, change in next changeset pushed

src/c2pa/c2pa.py
tmathern: Updating docs to match that, in next pushed changeset

src/c2pa/c2pa.py
tmathern: They are in the other PR

tests/test_unit_tests.py
tmathern: Not in Python, it mis-guesses . I have tests for that in the other PR too with the overloads!

src/c2pa/c2pa.py
tmathern: Noted, adding to other PR

PR #300, opened 2026-07-23
docs: Edit and cleanup pass

crandmck:
- Use `github-only` CSS class instead of inline `display:none` style to hide content in Docusaurus but still display in the GitHub web interface.


docs/native-resources-management.md
```diff
@@ -201,7 +198,7 @@ Every public method calls `_ensure_valid_state()` before doing any work, which r
 
 ## Ways to clean up
 
-### Context manager (`with` statement)
+### Using a context manager
```
tmathern: Please keep "(with statement)", as not everyone may know the name of the pattern (Context manager is the pattern, with is the keyword)

tmathern (changes requested):

tmathern (approved):

PR #303, opened 2026-07-28
feat: Binding for custom HTTP resolver (`with_resolver` on Context)

tmathern:
- Add bindings to be able to set custom HTTP resolved...
- ... which needed a way to retrieve the native malloc to free correctly (see _get_native_malloc), which is na interesting implementation detail
- Adds test examples of resolvers (I don't think this is ready for prime time in the examples fodler), with disclaimers those a re examples examplifying only one aspect of a resolver each, and multiple likely need to be combined when someone writes their own.

## Checklist


tests/http_resolver/__init__.py
```diff
@@ -0,0 +1 @@
+# Placeholder
```
ok-nick, line 1: 🤔

src/c2pa/c2pa.py
```diff
@@ -288,6 +292,32 @@ def _free_native_ptr(ptr):
...
+                except OSError:
+                    crt = ctypes.CDLL("msvcrt")
+            else:
+                crt = ctypes.CDLL(None)
+            malloc = crt.malloc
+            malloc.argtypes = [ctypes.c_size_t]
+            # The default c_int restype truncates 64-bit pointers.
+            malloc.restype = ctypes.c_void_p
+            ManagedResource._native_malloc = malloc
+        return ManagedResource._native_malloc
```
ok-nick, line 319: Hm, this doesn't seem too ideal. Maybe the C FFI should be accepting a free callback so the caller can provide their own allocation. We discussed this a little bit when Colin was adding the functionality to the C FFI. I suppose this is fine for now but something to think about.

tests/http_resolver/__init__.py
tmathern: Oh, that is for the moduling in Python, so various tools (who may not agree) don't complain. Unittest generally is happy with whatever, pytest is more easily confused, any linter loves to flag that as missing file otherwise.

PR #312, opened 2026-08-25
fix: Put a sentinel in the native thread local error slot

tmathern:
After a consuming native call fails, the Python side decides whether the handle was taken by reading the last native error and matching it against the pre-consume tags (`UntrackedPointer:`, `WrongPointerType:`). That slot is thread-local (`cimpl_error.rs`, `LAST_ERROR`) and sticky: nothing clears it between calls. A pooled worker thread therefore carries whatever error it last set, across unrelated tasks. A later failure that does not set its own error reads a tag from a task that finished long ago. THis leads to wrong decisions regarding what to release actually when.

This is especially problematic if an address was reused (which can happen under heavier load): the registry rejects an
address it no longer tracks, unless another thread has allocated a fresh tracked object at that exact address, in which case the stale free finds a live entry and destroys a different thread's object.

Repeated allocation and release of similarly sized objects is precisely the pattern that makes an allocator hand back an address it just freed. This leads either to bad references, or allocator-detected corruptions.

Put a sentinel in the error slot so the failing path can be identified, and to make sure no stale error is visible (which could lead to the wrong memory release decisions).

As a sidenote: On any consume failure where the verdict is not certain, this takes the consumed branch, meaning leaks could appear. Also, note this will disappear with v0.91.0 of c2pa-rs, which clarifies the consumption status of pointers (always consumed by native calls), make sure the leak situation can't happen anymore. It is still worthwhile to consider this fix, to ensure no stale errors are read by churning threads.

## Checklist


src/c2pa/c2pa.py
```diff
@@ -271,15 +271,17 @@ def __init__(self):
         self._pending_teardown = None
         record_owner_pid(self)
 
-    def _lock(self):
-        """Return this resource's operation lock.
+    def _state_lock(self):
+        """Return this resource's raw operation lock, with no side effects
+        beyond mutual exclusion.
```
ale-adobe: What does this mean? Mutual exclusion isn't a side effect in my mind.

src/c2pa/c2pa.py
```diff
@@ -306,6 +308,19 @@ def _lock(self):
                 pass
         return lock
 
+    @contextlib.contextmanager
+    def _lock(self):
+        """Hold this resource's operation lock its duration,
+        and mark this thread as inside a native-error section.
+
+        Never hold this across a native call that drives stream callbacks.
+        Those calls release the GIL and re-enter caller-supplied
```
ale-adobe: GIL?

src/c2pa/c2pa.py
```diff
@@ -306,6 +308,19 @@ def _lock(self):
...
+        code, which may call back into this API on another thread.
+        Only calls that touch no callbacks are serialized here.
```
ale-adobe: nit
```suggestion
        Only calls that don't touch callbacks are serialized here.
```

src/c2pa/c2pa.py
```diff
@@ -792,16 +844,47 @@ class C2paStream(ctypes.Structure):
...
+# in the native error slot.
+# Never a real handle: allocations are aligned, and the Python
+# layer only passes real handles or this constant to c2pa_free.
+_MARKER_ADDR = 1
+
+# Exact text the native lib writes for a failed free of _MARKER_ADDR.
+# Learned at import by _learn_sentinel_no_native_error_text().
+# The format is a native implementation detail,
+# so it is read back rather than hardcoded.
+_NO_NATIVE_ERROR_TEXT = None
```
ale-adobe: Makes me a bit more sense to me if it were named `NATIVE_NO_ERROR_TEXT`.

src/c2pa/c2pa.py
```diff
@@ -1135,6 +1267,38 @@ def _setup_function(func, argtypes, restype=None):
...
+    Runs on the importing thread; the text is a format constant, so the
+    learned value holds for every thread. Raises at import when the read
+    back text is empty, because the marker mechanism cannot work then.
+    """
+    _mark_sentinel_no_native_error()
+    raw = _lib.c2pa_error()
+    if not raw:
+        raise ImportError(
+            "c2pa native library did not report an error for a free of "
+            "an untracked pointer; the error-slot marker cannot work")
```
ale-adobe: For my understanding: without this, we could still technically operate, but it's just not necessarily thread-safe, right? (basically what we have now, before the PR lands).

src/c2pa/c2pa.py
```diff
@@ -1135,6 +1267,38 @@ def _setup_function(func, argtypes, restype=None):
...
+            "c2pa native library reported an empty error for a free of "
+            "an untracked pointer; the error-slot marker cannot work")
+    return text
+
+
+_NO_NATIVE_ERROR_TEXT = _learn_sentinel_no_native_error_text()
+assert "0x1" in _NO_NATIVE_ERROR_TEXT, (
+    "c2pa native library's untracked-pointer error text no longer "
+    "includes the planted address; the error-slot marker assumption "
+    "no longer holds")
```
ale-adobe: Any reason we don't assert this inside of the `_learn_sentinel_no_native_error_text()` call?

ale-adobe (approved):

src/c2pa/c2pa.py
tmathern: The Global Interpreter Lock (I will un-acronymize it in the docs). 
Python until Python 14 has an uber-lock around the interpreter. "In CPython, the global interpreter lock, or GIL, is a mutex that protects access to Python objects, preventing multiple threads from executing Python bytecodes at once." => Which means while the native code works, we are not under that lock anymore.
That: https://wiki.python.org/moin/GlobalInterpreterLock

src/c2pa/c2pa.py
tmathern: Yes, with care (currently we do have mostly that care).
Errors are not thread-safe at all form the C FFI layer (we use set_last_error, which is thread local). We also do not have a way to clear that slot before a call that may error, so you can read stale errors that were set by other calls (reading errors doesn't clear them).
This is the sentinel value to avoid hard coding our own version to set in an error slot to say "we actually got no error before this call", do the call, and check immediately afterwards. 
Also, since most code is aligned (except maybe embedded code?), the address 1 itself should never be allocated by an allocator, which is why I use it to trigger the error and get the error text the SDK sets (the whole error message + pointer in hex currently).

src/c2pa/c2pa.py
tmathern: What I was trying to say is that this accessor only takes the lock, in contrast to the sibling accessor below it, which also marks the thread as being inside a native-error section. Let me reword that.

src/c2pa/c2pa.py
tmathern: Let me fix that!

PR #314, opened 2026-08-25
fix: Protect native handles from concurrent access using locks

tmathern:

Sharing a `Reader`, `Builder`, `Signer` or `Context` across threads can free a native handle while another thread is still using it. This can cause segmentation faults and aborts as corruption is detected, especially under threaded environments (running as asyncio tasks) under heavy load.

This happens in the Python layer. For instance: In `Reader.json()`, `self._handle` was read first, and then the native call was made using the value that was read. Those are two separate steps, so a `close()` on another thread could free the handle in between. The freed address can then be reused (which happens quickly under churn, and tests also noticed it when ran in repeated loops): a verification workload allocates and releases readers continuously, so the allocator returns an address it released moments earlier and a new reader registers there. The stale call now finds the address tracked (and the recorded type is even correct), so native registry checks pass and the call operates on an object owned by another thread. The native library only receives an address (pointers), and can't know the value it got stopped being logcally valid and usuable on the other side of the fence. (This is a variant of ABA problem).

### What this PR does

Each managed resource gains a reentrant operation lock (reentrant so a thread would not be capable to block on itself), held only for short regions that read and write handle state. It is reentrant because a finalizer can run at any point, including inside a region the same thread already holds, and because a consuming call tears the handle down from inside the locked region.

The lock is never held across a native call that runs caller-supplied Python code. Those calls release the interpreter lock and re-enter the caller's callbacks, which may call back into this API on another thread, so holding a lock across them could deadlock. They are counted as in flight instead. A teardown arriving while a call is in flight records its intent rather than freeing, and whichever caller leaves the in-flight region last performs the free. The resource is marked closed as soon as the intent is recorded, so a caller that closed it cannot keep using it while the free is pending.

Important: The recorded intent only ever moves from "free it" to "do not free it", never back. Ownership does not return once a consuming call has handed the pointer to the native library, so a later teardown cannot restore a right to free that was already given away.

`Reader.with_fragment()` consumes and replaces the handle, so it serializes against itself under its own lock.

Behavior in a forked child is unchanged. Acquiring the operation lock there raises rather than waiting, because a child inherits the lock in whatever state it had at `fork()` and the thread holding it does not exist to release it.

The [test test_no_conflicting_lock_acquisition_order](https://github.com/contentauth/c2pa-python/blob/fce5327fe9fcfbbac787ff0d43ed5e446cc6dacd/tests/test_unit_tests.py#L9685-L9790) enforces lock acquistion order also (code analysis) to make sure we're not set up for deadlocks.

### Scope
No change to the native library and no change to how callers dispatch work.
The gap closed is between the attribute read and the call, which is Python-handled (~handover between Python and native) so it is addressed here: the steps that have to be together (in a locked/critical section) are in Python, so Python needs to handle it. This also means that this fix does not depend on a new native lib release: once the Python fix is released, since the fix is in Python, the hole/issue is closed.

### Traced example of what could go wrong before (details)

This example traces one case where, without locks, we could hit SIGSEV or SIGABORT. The error depends on what detects it (auto-detect causes sigabort, hard crash sigsev). Note this scenario traces what happens before the fix, to (try to) explain the encountered issue. This could happen with any native resource.

The following sequence is for a same Reader:

1. Thread A calls `reader.json()`. `_ensure_valid_state()` passes, because the state is `ACTIVE` and the handle is set. Execution continues to `_lib.c2pa_reader_json(self._handle)`.
2. That line is two separate steps: `self._handle` is read, and then the call is made using the value that was read. Nothing holds those two steps together, so a thread switch between them is actually permitted.
3. Thread B calls `reader.close()` in that interval. It reaches `_teardown(free_handle=True)`, which sets the state to `CLOSED`, nulls `self._handle`, and passes the old value to `c2pa_free`. The handle is freed. Nulling the attribute does not reach thread A, which already read the address.
4. Another reader is constructed, on any thread. There can be one per asset, so same-sized objects are allocated and released continuously and the allocator returns the address it released moments earlier. The new reader registers at that address.
5. Thread A now makes its call with the old address. The registry finds it tracked, (and the recorded type is `Reader`), so the registry checks pass. The call operates on a reader owned by another thread. Depending on how far the reuse has progressed that is a wrong result, a segmentation fault, or an allocator abort (and we will get the matching error).

## Checklist


src/c2pa/c2pa.py
```diff
@@ -264,8 +266,80 @@ def _init_attrs(self):
...
+            with self._lock():
+                self._inflight -= 1
+                pending = (self._pending_teardown
+                           if self._inflight == 0 else None)
+                if pending is not None:
+                    self._pending_teardown = None
+            # Released the lock before the free:
+            # _teardown takes it again, and keeping the two acquisitions
+            # separate means the counter update is never held across
+            # the release work.
```
ale-adobe: 👍

src/c2pa/c2pa.py
```diff
@@ -321,22 +395,51 @@ def _safe_release(self):
     def _teardown(self, free_handle: bool):
         """Close the object: run _release, optionally free the handle, null it.
         free_handle=False (consumed) frees nothing, the new owner needs to free.
+
+        Holds the operation lock so the free cannot happen between another
+        thread's state check and its use of the handle in a native call.
+
+        The forked-child case is handled before the lock is taken, because
+        _lock() refuses in a child: this path has to finish rather than report
```
ale-adobe:
```suggestion
        _lock() raises in a child: this path has to finish rather than report
```

docs/native-resources-management.md
```diff
@@ -106,15 +106,27 @@ Therefore, the managed resources have the following principles:
 
 ### Double-free risk mitigations
 
-Three distinct risks. Two have a mechanism in this layer; the third is the caller's to synchronize:
+Three distinct risks, each with its own mechanism in this layer:
 
 | Hazard | Covered by | How |
 | --- | --- | --- |
-| Freeing a pointer a consuming call already took (single flow) | `_swap_handle` / `_teardown(free_handle=False)` triage | The consumed pointer is abandoned, never freed. The retained-vs-consumed decision reads the native error tag (`UntrackedPointer:` / `WrongPointerType:` mean not taken). |
+| Freeing a pointer a consuming call already took (single flow) | `_swap_handle` / `_teardown(free_handle=False)` triage | The consumed pointer is abandoned, never freed. The retained-vs-consumed decision reads the native error tag (`UntrackedPointer:` / `WrongPointerType:` / `NullParameter:` / `InvalidBufferSize:` mean not taken). |
```
ale-adobe: What was the context for the addition of these two error tags? Just completeness?

docs/native-resources-management.md
```diff
@@ -106,15 +106,27 @@ Therefore, the managed resources have the following principles:
...
+
+`_op_lock` is an `RLock` rather than a plain `Lock` for two reasons specific to this code. First, a finalizer (`__del__`) can run at any bytecode boundary — including one in the middle of a method that has already acquired the lock on this same thread — so `__del__` calling back into locked code must not deadlock against itself. Second, a consuming call tears the handle down from inside the locked region it is already holding: `_teardown()` is called while `_op_lock` is held, and it needs to acquire the same lock again rather than re-entering as a different, blocked acquisition. `_lock()` returns it, except in a forked child: there it raises `C2paError` immediately rather than blocking, because the thread that might hold the lock at fork time does not exist in the child to release it, and waiting on it would hang forever (see [Fork safety](#fork-safety)).
```
ale-adobe, line 238: This was very helpful to include, thank you.

docs/native-resources-management.md
tmathern: Yes, and I suspect they may be actually hit. Initially I added only those which were relevant for reported leaks, but those two might also be in this context to identify ownerships of pointers/handles.

mstattma (commented):
EDIT: sorry review was an agent mistake

gpeacock (approved):
I understand what you are doing here and why. This is ok to fix the current issue, but we really need to change the c_ffi tracking model so this isn't an issue. I know how to do that. 
Currently we pass around the actual allocated addresses of Rust allocated objects. I can switch that to always passing an incremental token, so each allocation gets a unique identifier. Eventually that would wrap around, but we would still have the type checks so that the chance this could happen become miniscule.

PR #320, opened 2026-08-30
Drop wheel, setuptools and pytest from runtime dependencies

fwittreverce:
`wheel`, `setuptools` and `pytest` are declared in `[project.dependencies]`, so
every consumer installs three build/test packages into production environments.
None of the three is imported anywhere under `src/`.

`pytest` is re-declared as a PEP 735 `[dependency-groups] dev` entry rather than
dropped outright, so the manifest still names the test dependency — `uv sync`
and `pip install --group dev` pick it up — while keeping it out of the published
package metadata, which `[project.optional-dependencies]` would not do. The
bound follows `requirements-dev.txt` (`>=8.1.0`) rather than the `>=7.4.0` the
runtime entry carried; nothing installs the older one today.

### Evidence

Every import in the installed package, on `main` (0.37.9):

| Module | Imports |
| --- | --- |
| `src/c2pa/c2pa.py` | stdlib only |
| `src/c2pa/lib.py` | stdlib only |
| `src/c2pa/__init__.py` | stdlib + own module |
| `src/c2pa/build.py` | `requests` (line 16), `toml` (line 97, lazy) |

So `requests` and `toml` are genuine runtime dependencies — `build.py` is the
installed `download-artifacts` console script — and this PR leaves both alone.
`wheel`, `setuptools` and `pytest` are imported by nothing.

This repo already classifies all three correctly in three other places:

* **`[build-system] requires`** already lists `setuptools>=68.0.0` and `wheel`,
  so the build has what it needs; the runtime entries are duplicates.
* **`requirements-dev.txt`** lists `wheel` and `setuptools` under
  `# Build dependencies` and `pytest` under `# Testing dependencies`.
* **`.github/workflows/build.yml`** installs pytest explicitly
  (`pip install pytest`, lines 285 and 377), so CI does not rely on the runtime
  declaration either.

The change is therefore a no-op for this repo's own build and test paths.

### Why it is worth doing

**They reach production images.** In our worker, `uv export --frozen --no-dev` —
the resolver's production set — lists all three. `setuptools` in particular has
a CVE history, so a security scan has to triage findings for packages the
application never imports.

**They mask missing test dependencies downstream, silently.** This is the one
that cost us time. Our own test extra was never actually installed — `uv sync`
does not install extras — and nobody noticed for seven weeks, because `pytest`
arrived through this dependency chain anyway. Our suite ran on a package no
manifest of ours declared, at a version nobody chose, and a fix we shipped in
that window was inert the whole time. A dependency audit is what eventually
flagged it.

### One question, deliberately not in this diff

`cryptography` is also unimported under `src/` — it appears only in `examples/`
and `tests/`, and your own `requirements.txt` says `# only used in the training
example`. It looked like your call rather than mine: dropping it would stop
`pip install c2pa-python` from giving a reader everything the signing examples
need. Happy to extend the PR if you would rather it went too.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,11 +27,6 @@ dependencies = [
     "requests>=2.0.0"
 ]
 
-[dependency-groups]
-dev = [
-    "pytest>=8.1.0"
-]
-
 [project.scripts]
 download-artifacts = "c2pa.build:download_artifacts"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,10 +22,9 @@ maintainers = [
 ]
 urls = {homepage = "https://contentauthenticity.org", repository = "https://github.com/contentauth/c2pa-python"}
 dependencies = [
-    "wheel>=0.41.2",
-    "setuptools>=68.0.0",
+    # `toml` and `requests` are read at runtime by the `download-artifacts`
+    # console script (src/c2pa/build.py), so they stay.
     "toml>=0.10.2",
-    "pytest>=7.4.0",
     "cryptography>=41.0.0",
     "requests>=2.0.0"
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,11 +22,14 @@ maintainers = [
 ]
 urls = {homepage = "https://contentauthenticity.org", repository = "https://github.com/contentauth/c2pa-python"}
 dependencies = [
-    # `toml` and `requests` are read at runtime by the `download-artifacts`
-    # console script (src/c2pa/build.py), so they stay.
     "toml>=0.10.2",
     "cryptography>=41.0.0",
     "requests>=2.0.0"
+]
+
+[dependency-groups]
+dev = [
+    "pytest>=8.1.0"
 ]
 
 [project.scripts]


tmathern:
For further review and consideration, please make sure to sign the Adobe CLA. You will likely need to close and reopen the PR for the job to pass.

fwittreverce:
@tmathern fixed.

pyproject.toml
```diff
@@ -22,14 +22,16 @@ maintainers = [
...
-    "setuptools>=68.0.0",
     "toml>=0.10.2",
-    "pytest>=7.4.0",
     "cryptography>=41.0.0",
     "requests>=2.0.0"
 ]
 
+[dependency-groups]
+dev = [
+    "pytest>=8.1.0"
```
tmathern: There is a pytest in:
https://github.com/contentauth/c2pa-python/blob/main/requirements-dev.txt#L8. What do you think of completely removing it here, and keeping the one in requirements-dev.txt only, so they couldn't drift?

fwittreverce: Happy to — removing it outright is the smaller diff, and it makes this PR purely
subtractive. Nothing in the repo reads a `[dependency-groups]` entry today: the dev
path is `pip install -r requirements-dev.txt` everywhere (`make install-deps`, and
`build.yml` lines 52, 90/93, 164/167, 490/492). The group was defensive rather than
load-bearing — the point of the PR is the published metadata, which is fixed either
way. I'll push that.

One observation, offered rather than requested: the drift here is structural rather
than accidental, so removing my line removes one instance of it, not the mechanism.
Today `pytest` reads `>=7.4.0` in `pyproject.toml` against `>=8.1.0` in
`requirements-dev.txt`, and `cryptography` reads `>=41.0.0` in `[project.dependencies]`
against `>=47.0.0` in both requirements files. This PR settles the first pair; the
second stays.

If you ever wanted that impossible rather than merely absent, the standards have since
caught up with everything these files are doing:

| what it declares | standard home |
| --- | --- |
| published runtime deps | `[project.dependencies]` — already there |
| build backend deps | `[build-system] requires` — already there; the `wheel` / `setuptools` / `toml` lines in `requirements-dev.txt` duplicate it |
| dev sets: test, lint, docs | `[dependency-groups]` — PEP 735, final 2024-10-10; `pip install --group <name>` since pip 25.1 |
| a pinned environment | `pylock.toml` — PEP 751, final 2025-03-31 |

which for this repo is roughly:

```toml
[dependency-groups]
test = ["pytest>=8.1.0", "cryptography>=47.0.0"]
lint = ["autopep8>=2.3.0", "flake8==7.3.0"]
docs = ["Sphinx>=7.3.0", "sphinx-autoapi>=3.0.0", "myst-parser>=2.0.0", "furo>=2024.0.0"]
dev  = [{include-group = "test"}, {include-group = "lint"}, {include-group = "docs"}]
```

with `make install-deps` becoming `pip install --group dev`. Every bound would then
exist exactly once, and a group cannot drift from itself.

Two caveats I would rather name than have you discover: it needs pip >= 25.1 on every
path that installs, and it is not a small diff — `Makefile`, `build.yml`,
`build-wheel.yml`, `publish-docs.yml`, `docs/project-contributions.md` (eight
references) and the perf Dockerfiles all name the requirements files today.
`requirements.txt` would need a decision rather than a translation: its single line is
an example dependency that is also published as a runtime dep, which is the same
question I left open at the bottom of this PR.

Your call entirely, and clearly not this PR — I am happy to open it separately if you
want it, and equally happy to leave it be.

tmathern:
Builds don't trigger on PRs from forks, so running tests there: https://github.com/contentauth/c2pa-python/pull/321

tmathern (approved):

tmathern:
LGTM, thanks for the contribution!

ale-adobe (approved):

PR #323, opened 2026-09-10
fix(deps): drop cryptography from runtime dependencies

AlexanderKomarov:

Not used by the runtime bindings — all crypto happens in the native
c2pa-rs library. Only the tests/examples need it (callback signers), and
it is already declared in requirements-dev.txt. Drop it from the runtime
deps and install requirements-dev.txt in the built-wheel CI jobs.

## Checklist
- [ ] Any `TO DO` items (or similar) have been entered as GitHub issues — N/A (no TODOs)


.github/workflows/build.yml
```diff
@@ -272,7 +272,8 @@ jobs:
       - name: Install wheel for testing
         run: |
           source venv/bin/activate
-          pip install dist/c2pa_python-*.whl
+          # `[test]` extra pulls cryptography for the callback-signer tests.
```
tmathern: Instead of mirroring requirements-dev.txt, why not use requirements-dev.txt?

AlexanderKomarov: good call, fixed

PR #325, opened 2026-09-14
Preflight: validate against c2pa-rs 0.91.0-rc.3

scouten-adobe:
## Summary

Now that #326 has split out the reusable tooling and general test fixes, this PR is just the RC-specific part: it points `c2pa-rs-preflight-ref.txt` (added in #326) at a c2pa-rs release candidate tag, so `test-c2pa-rs-source-build.yml` builds the native library from that git tag and runs the full test suite against it on every push.

**Depends on #326** — this PR's base is that branch, and it should be retargeted to `main` once #326 merges.

## Update: now tracking rc.3

Originally pointed at `c2pa-rc-v0.91.0-rc.1`, which reproduced 8 failures per platform (`'Valid' != 'Trusted'` plus one SDK-version string mismatch) caused by `trust.trust_anchors`/`trust.trust_config` moving under a new `trust.anchors[]` array: the deprecated `Settings::from_string()` thread-local path (what `c2pa_load_settings`/our `load_settings()` use) wasn't migrating the legacy fields into the new schema.

Between rc.1 and rc.3, c2pa-rs fixed exactly that — [contentauth/c2pa-rs#2663](https://github.com/contentauth/c2pa-rs/issues/2663), landed with a regression test in `sdk/src/settings/mod.rs`. Rebuilt locally and reran the full suite against `c2pa-rc-v0.91.0-rc.3`: all 7 trust-config tests now pass. CI confirms the same across all 5 platforms (macOS, Linux x64/arm64, Windows x64/arm64) -- the only remaining failure is the expected `test_sdk_version` string check, which just compares against whatever's in the currently-pinned `c2pa-native-version.txt` (0.90.22) and isn't a real issue.

## Test plan

- [x] Local build from `c2pa-rc-v0.91.0-rc.3` source (macOS arm64) + full unit test suite: 443/444 pass (the 1 failure is the expected version-string check)
- [x] CI (`resolve-ref` + `tests-unix`/`tests-windows` matrix): same result on all 5 platforms
- [ ] Once c2pa-rs 0.91.0 is fully released with prebuilt artifacts, this PR's job is done -- do the real version bump (`c2pa-native-version.txt`, plus migrate `examples/read.py`'s trust config to the new schema) as its own PR, and delete `c2pa-rs-preflight-ref.txt`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/build.yml
```diff
@@ -98,13 +98,24 @@ jobs:
           mkdir -p src/c2pa/libs
           rm -rf dist/* build/*
 
-      - name: Download native artifacts
+      # c2pa-rs ${{ needs.read-version.outputs.c2pa-native-version }} is a
```
tmathern, line 101: Instead of reverting, maybe put it behind a label or a manually triggered workflow that takes the version number as param? So you don't need to repeat the workflow exercise for every pre-release

tests/test_unit_tests.py
```diff
@@ -3444,7 +3524,7 @@ def test_builder_add_multiple_ingredients(self):
         # Test adding another ingredient
         ingredient_json = '{"test": "ingredient2"}'
         with open(self.testPath2, 'rb') as f:
-            builder.add_ingredient(ingredient_json, "image/png", f)
+            builder.add_ingredient(ingredient_json, "image/jpeg", f)
```
tmathern, line 3527: That is weird. Was the format in the tests wrong? Why is it changing here?

tmathern (commented):
Do the examples in https://github.com/contentauth/c2pa-python/tree/main/examples also need updates?

.github/workflows/build.yml
scouten-adobe: Good idea, and even better than I first implemented it. Done in 365e6dd + cc53a30: build.yml and c2pa-native-version.txt are back to the real 0.90.19 release/download flow, and the source-build logic now lives in test-c2pa-rs-source-build.yml as a standalone workflow.

I first made it workflow_dispatch-only, but that doesn't give a PR real CI signal while iterating on adaptation commits -- you'd have to remember to re-dispatch by hand on every push, so reviewers can't just trust the PR's check list. So it now also reruns automatically: commit the target ref to a new c2pa-rs-preflight-ref.txt file, and the workflow picks it up and runs the full matrix on every push to that PR (same as any other check), falling back to workflow_dispatch's ref input for true one-offs with nothing committed. This PR is now exercising that path against 0.91.0-rc.1.

tests/test_unit_tests.py
scouten-adobe: Yes, it was a pre-existing bug in the test, just a latent one -- testPath2 is A.jpg (an actual JPEG), but four tests declared it as "image/png" when adding it as an ingredient. Older c2pa-rs didn't validate ingredient bytes against the declared MIME type at add-time, so the mismatch went unnoticed. 0.91.0-rc.1 now does validate it and errored on the mismatch, which is what surfaced the bug. Fixed by correcting the declared type to match the actual bytes -- verified it still passes against the currently-pinned 0.90.19 too.

scouten-adobe:
Re: examples/ -- examples/read.py uses the same old \`trust.trust_anchors\` flat schema as our test fixture did, so it'll need the same migration to \`trust.anchors[]\` whenever we actually bump to 0.91.0. Not needed for this PR though, since we're keeping the pin at 0.90.19 and it works fine there today.

tmathern (approved):

